PR #11, opened 2024-10-16
Update default embedding model

fzliu:


thomas0809:
We cannot change the default model here because it causes backward incompatibility. In other words, if a user does not set model explicitly and uses the default model to build the index, after this change, they will be using a different model and get embeddings that are not compatible with their index.

`model` is a required argument. Due to historical reason, it defaults to `None` and a warning will be thrown if `model` is not set. So one TODO is to really make `model` required, throw error if not set, and remove default model. I think it's about time to do that.

PR #15, opened 2024-11-02
[Multimodal] Revise dependency versions

kenhong:
- Revise dependencies to allow for wider compatibility
- Remove support for Python 3.8 and lower (end-of-life)


thomas0809 (approved):
Looks great! Thanks!

PR #19, opened 2024-11-23
[Functionality] Add API params for supporting output quantization

kenhong:
- Add optional params `output_dtype` and `output_dimension` to `embed`
- Bump version
- Unit tests

### DIFF
--- a/voyageai/api_resources/embedding.py
+++ b/voyageai/api_resources/embedding.py
@@ -29,7 +29,7 @@ class Embedding(APIResource):
                 # If an engine isn't using this optimization, don't do anything
                 if type(data["embedding"]) == str:
                     data["embedding"] = np.frombuffer(
-                        base64.b64decode(data["embedding"]), dtype=map_output_dtype(kwargs["output_dtype"])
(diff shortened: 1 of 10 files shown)


voyageai/api_resources/embedding.py
```diff
@@ -28,7 +29,7 @@ def create(cls, *args, **kwargs):
                 # If an engine isn't using this optimization, don't do anything
                 if type(data["embedding"]) == str:
                     data["embedding"] = np.frombuffer(
-                        base64.b64decode(data["embedding"]), dtype="float32"
+                        base64.b64decode(data["embedding"]), dtype=map_output_dtype(kwargs.get("output_dtype", "float"))
```
thomas0809, line 32: I don't like the name `map_output_dtype`, which is not meaningful by itself.

I suggest to move the entire `np.frombuffer(***)` part to a utility function, such as `decode_base64_embedding(data: str, dtype: str | None)`.

I am not sure whether we should put it in voyageai.util, as no other code is using it. So I would keep the function in this file, but it's okay to me if you want to move it to util.

thomas0809 (approved):

PR #24, opened 2025-04-25
Allow python 3.13+

Askir:
Could we try if this passes the test suite, I don't think you need version <3.13 and people would like to use newer python versions :D


kenhong:
[Addressed in v0.3.3](https://github.com/voyage-ai/voyageai-python/releases/tag/v0.3.3)

PR #26, opened 2025-07-01
[Dependency] Add support for Python 3.13

kenhong:
- Add support for Python 3.13, revise dependency versions
- Revise test case to reflect updated max batch size for embeddings


thomas0809 (approved):
LGTM

PR #49, opened 2026-02-04
Local model (voyage-4-nano) support

fzowl:
Seamless integration of local model(s)


fzowl:

PR #46, opened 2026-01-27
Python 3.14 support

fzowl:

PR #50, opened 2026-03-21
Migrate to Pydantic v2 APIs for Python 3.14 support

fzowl:
The previous fix (#46) swapped the pydantic import order but still fell back to the broken pydantic.v1 path since Extra was removed in v2.

- Remove pydantic.v1 fallback and Extra import
- Replace Extra.forbid with string "forbid"
- Replace min_items with min_length (v2 rename)
- Replace .parse_obj() with .model_validate()
- Replace .dict() with .model_dump()
- Bump pydantic lower bound to >=2.7.4 (already required transitively by langchain-text-splitters via langchain-core)

Fixes #36

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "voyageai"
-version = "0.3.8"
+version = "0.3.9"
 description = ""
 authors = ["Yujie Qian <yujieq@voyageai.com>"]
 readme = "README.md"
@@ -16,7 +16,7 @@ numpy = [
 ]
 aiolimiter = "*"
 pillow = "*"
-pydantic = ">=1.10.8"
+pydantic = ">=2.7.4"
 tokenizers = ">=0.14.0"
 langchain-text-splitters = ">=0.3.8"
 ffmpeg-python = "*"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "voyageai"
-version = "0.3.9"
+version = "0.4.0"
 description = ""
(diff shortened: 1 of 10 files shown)


abgros:
@fzowl what's the status of this PR?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "voyageai"
-version = "0.3.8"
+version = "0.3.9"
```
jasonqiu95: Going from pydantic >=1.10.8 to pydantic >=2.7.4 is a breaking change for any downstream consumer pinning v1. So it should be 0.4.0. We should also note this clearly that we switched to pydantic v2 in the release note.

tests/test_client_async.py
```diff
@@ -1,9 +1,16 @@
...
+
 import pytest
 import voyageai
 import voyageai.error as error
 from voyageai.chunking import default_chunk_fn
 
+requires_api_key = pytest.mark.skipif(
+    not os.environ.get("VOYAGE_API_KEY"), reason="VOYAGE_API_KEY not set"
+)
+
```
jasonqiu95, line 11: ~~this change seems unrelated could you move it to another PR and explain why it's needed?~~

nvm I figured out this is to make your fork's CI green right? Also I trust that the main repo's CI run will run these tests with an actual key right?

jasonqiu95 (approved):

pyproject.toml
fzowl: Good call — bumped to 0.4.0.

tests/test_client_async.py
fzowl: Yeah, exactly!

fzowl:

Issue #42, opened 2025-12-08
Circular Import Error: `AttributeError: partially initialized module 'voyageai' has no attribute 'error'`


chonknick:
## Description

The `voyageai` package (v0.3.5) has a circular import bug that causes `AttributeError` in certain import scenarios, particularly when used with lazy import systems.

## Error Message

```
AttributeError: partially initialized module 'voyageai' has no attribute 'error'
(most likely due to a circular import)
```

## Root Cause

The circular import occurs due to the following import chain:

1. `voyageai/__init__.py` imports from `voyageai/embeddings_utils.py`
2. `voyageai/embeddings_utils.py` imports `voyageai` (line 12)
3. `voyageai/embeddings_utils.py` immediately tries to use `voyageai.error.RateLimitError` (line 22)
4. However, `voyageai/__init__.py` never explicitly imports the `error` module
5. Result: `voyageai.error` doesn't exist when `embeddings_utils.py` tries to access it

## Code Analysis

**File: `voyageai/embeddings_utils.py`**

```python
# Line 12
import voyageai

# Lines 19-23
@retry(
    wait=wait_random_exponential(min=1, max=20),
    stop=stop_after_attempt(6),
    retry=retry_if_exception_type(voyageai.error.RateLimitError),  # ❌ error module not yet loaded!
)
def _get_embeddings(...):
    ...
```

**File: `voyageai/__init__.py`**

The `__init__.py` imports from `embeddings_utils` but never explicitly imports `voyageai.error`:                                                                                                                  ```python                                                                                                
from voyageai.embeddings_utils import (                                                                      
get_embedding,                                                                                           
get_embeddings,                                                                                          
...                                                                                                  
)
# ❌ Missing: from voyageai import error                                                                 
```


## Reproduction                                                                                                                                                                                                   The bug is intermittent and depends on:
- Import timing and order                                                                                
- Python version
- Whether `voyageai` was previously imported in the process                                             
- Lazy import patterns (common in ML libraries)

### Minimal Reproduction
```python                                                                                                
def lazy_import_test():
    """Simulates lazy import pattern that triggers the bug"""
    try:                                                                                                         
        import voyageai
        # This may fail with AttributeError                                                                      
        _ = voyageai.error.RateLimitError
        print("✓ Success")
    except AttributeError as e:
        print(f"✗ Error: {e}")                                                                                                        
        lazy_import_test()
```

## Impact                                                                                                                                                                                                         This bug affects:
[...]

PR #51, opened 2026-03-24
Fix/circular import error 42

fzowl:
Resolving https://github.com/voyage-ai/voyageai-python/issues/42


fzowl:

PR #54, opened 2026-05-19
Add auto-chunking parameters to contextualized_embed

jasonqiu95:
## Summary
- Add `enable_auto_chunking`, `chunk_size`, `chunk_overlap` to `Client.contextualized_embed` and `AsyncClient.contextualized_embed`, forwarded to `/contextualizedembeddings` for server-side chunking.
- Widen `inputs` to `Union[List[List[str]], List[str]]` and normalize flat input for `input_type="query"` or auto-chunking; otherwise raise the same error the gateway returns.
- Parse `chunker_version` and per-chunk `text` from the response; surface them on `ContextualizedEmbeddingsObject`.
- Existing `chunk_fn` clients are unaffected (no warnings, same behavior); `chunk_fn` is mutually exclusive with `enable_auto_chunking=True`.

Server-side validation lives at `voyageai-backend/src/python/common/core/api/api_gateway.py:259-321`; the client-side helper in `voyageai/chunking.py` mirrors those rules so users see consistent errors.

Design reference: voyage-context-4 launch eng plan, P1.6 SDK updates.

## Test plan
- [x] `pytest tests/test_client.py tests/test_client_async.py -k "auto_chunk or chunk_overlap or chunk_size_requires or chunk_fn_and_auto or query_flat or auto_chunking" -v` (18 pass — 9 sync + 9 async)
- [x] Full collection still works: `pytest tests/test_client.py tests/test_client_async.py --collect-only` (59 tests)
- [x] Manual smoke against staging once the gateway `enable_auto_chunking` feature flag is on:
  ```python
  vo = voyageai.Client()
  r = vo.contextualized_embed(
      inputs=["Long doc text..."],
      model="voyage-context-3",
      input_type="document",
      enable_auto_chunking=True,
      chunk_size=512,
      chunk_overlap=64,
  )
  assert r.chunker_version is not None
  assert r.chunk_texts and r.chunk_texts[0]
  ```
- [ ] Confirm legacy `chunk_fn` path still works end-to-end against staging.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1,10 +1,13 @@
 import importlib.metadata
+import re
 from typing import List
+from unittest.mock import patch
 
 import pytest
 import voyageai
 import voyageai.error as error
 from langchain_text_splitters import RecursiveCharacterTextSplitter
+from voyageai.api_resources.response import VoyageResponse
 from voyageai.chunking import default_chunk_fn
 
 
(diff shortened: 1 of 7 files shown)


fzowl (dismissed):
Nice feature addition — the validation function is well-structured, sync/async parity is clean, and the normalized_inputs / request_inputs split is a good fix for the retry-mutation bug. A few things I'd like to see addressed before merging, mostly around the silent failure path and error handling in response parsing. Details in the inline comments.

voyageai/object/contextualized_embeddings.py
```diff
@@ -20,25 +20,50 @@ def __init__(
...
+            if populated and len(populated) != len(server_texts_per_doc):
+                raise ValueError(
+                    "response returned chunk texts for some documents but not others; "
+                    "expected all-or-nothing"
                 )
+            if populated:
+                self.chunk_texts = populated
+
         self.total_tokens += response.usage.total_tokens
+        self.chunker_version = response.get("chunker_version")
```
fzowl, line 70: This is the one I'd most like to see addressed: if a user sets `enable_auto_chunking=True` but the server doesn't support it yet (feature flag off, older API), everything silently returns `chunker_version=None` and `chunk_texts=None` — no way to tell it didn't work.

Could we thread an `auto_chunking_requested` flag into the object and `warnings.warn()` here when the version is missing?

voyageai/object/contextualized_embeddings.py
```diff
@@ -20,25 +20,50 @@ def __init__(
...
-                        embeddings=embeddings,
+                per_doc_texts = [embd.get("text") for embd in d.data]
+                if all(t is not None for t in per_doc_texts):
+                    server_texts_per_doc.append(per_doc_texts)
+                    result_chunk_texts = per_doc_texts
+                elif any(t is not None for t in per_doc_texts):
+                    raise ValueError(
+                        f"inputs[{i}] returned a partial set of chunk texts; "
+                        "expected text on every chunk or none"
                     )
```
fzowl, line 46: A couple of thoughts on using `ValueError` here:

- This is a malformed *server response*, not bad user input — a `voyageai.error` subclass might be less surprising for callers who `except ValueError` their own validation.
- The embeddings are still valid at this point. Maybe degrade gracefully (warn + set `chunk_texts=None` for this doc) so the user keeps what they paid for?
- Also watch out — this raises mid-loop, leaving `self.results` partially populated. Might be safer to finish the loop first, then validate.

voyageai/object/contextualized_embeddings.py
```diff
@@ -20,25 +20,50 @@ def __init__(
...
+                result_chunk_texts = client_chunk_texts[i]
             else:
-                self.results.append(
-                    ContextualizedEmbeddingsResult(
-                        index=i,
-                        embeddings=embeddings,
+                per_doc_texts = [embd.get("text") for embd in d.data]
+                if all(t is not None for t in per_doc_texts):
+                    server_texts_per_doc.append(per_doc_texts)
+                    result_chunk_texts = per_doc_texts
```
fzowl, line 41: Minor: `.get("text")` returns `None` both when the key is absent and when the server sends `"text": null`. Might be worth using `"text" in embd` if you ever need to tell the difference (e.g., null = server-side error vs absent = feature not available).

voyageai/object/contextualized_embeddings.py
```diff
@@ -20,25 +20,50 @@ def __init__(
...
+                    chunk_texts=result_chunk_texts,
+                )
+            )
+
+        if client_chunk_texts is None and server_texts_per_doc:
+            populated = [t for t in server_texts_per_doc if t is not None]
+            if populated and len(populated) != len(server_texts_per_doc):
+                raise ValueError(
+                    "response returned chunk texts for some documents but not others; "
+                    "expected all-or-nothing"
```
fzowl, line 64: Since `update()` accumulates `results` and `total_tokens`, it looks like it's meant to be called more than once (batched responses?). But `self.chunk_texts` and `self.chunker_version` (line 69) get replaced rather than extended. Worth either switching to `extend()` or adding a note that `update()` should only be called once when server-returned texts are present.

voyageai/chunking.py
```diff
@@ -29,6 +29,70 @@ def apply_chunking(
...
+        and chunk_overlap >= chunk_size
+    ):
+        raise ValueError(
+            f"chunk_overlap ({chunk_overlap}) must be less than chunk_size ({chunk_size})"
+        )
+    if chunk_size is not None and chunk_size < 1:
+        raise ValueError("chunk_size must be greater than or equal to 1")
+    if chunk_overlap is not None and chunk_overlap < 0:
+        raise ValueError("chunk_overlap must be greater than or equal to 0")
+
```
fzowl, line 61: Right now `chunk_overlap=32` without `chunk_size` passes validation and gets forwarded to the server. Is that intentional (server has a default chunk_size)? If so, a quick comment would help. Otherwise:
```python
if chunk_overlap is not None and chunk_size is None:
    raise ValueError("chunk_overlap requires chunk_size")
```

voyageai/chunking.py
```diff
@@ -29,6 +29,70 @@ def apply_chunking(
...
+    if isinstance(inputs, list) and all(isinstance(s, str) for s in inputs):
+        if input_type != "query" and not enable_auto_chunking:
```
fzowl, line 63: Small gotcha: `all(isinstance(s, str) for s in [])` is `True`, so `inputs=[]` sails through and hits the API. An early `if not inputs: raise ValueError(...)` would give a clearer error.

voyageai/client.py
```diff
@@ -93,24 +96,38 @@ def embed(
...
+            input_type=input_type,
+            chunk_fn=chunk_fn,
+            enable_auto_chunking=enable_auto_chunking,
+            chunk_size=chunk_size,
+            chunk_overlap=chunk_overlap,
+        )
+
         response = None
         for attempt in self._make_retry_controller():
             with attempt:
```
fzowl, line 124: Nit: since `apply_chunking` is pure and `normalized_inputs` doesn't change, you could move the chunking before the retry loop to avoid re-doing it on each attempt. Also makes `request_inputs` available outside the loop without relying on the `response is None` guard.

(Same in `client_async.py`.)

tests/test_client.py
```diff
@@ -291,6 +294,218 @@ def test_client_contextualized_embed_output_dtype(self):
...
+                model=self.context_embed_model,
+                input_type="query",
+            )
+
+        assert captured["inputs"] == [["hello"]]
+        assert "enable_auto_chunking" not in captured
+        assert result.chunker_version is None
+        assert result.chunk_texts is None
+
     """
```
fzowl, line 548: Nice coverage! A few more cases that'd be good to have:
- `chunk_size=0` and `chunk_overlap=-1` (boundary checks on lines 62-65 of chunking.py)
- Server returning partial `text` within a doc / across docs (the two `ValueError` paths in `contextualized_embeddings.py`)
- Asserting `results[1].chunk_texts` too — currently only `results[0]` is checked

fzowl (changes requested):
Great work overall! The validation logic is solid and the normalized_inputs / request_inputs split nicely fixes the retry-mutation issue. Left a few comments — mostly around making sure users get clear feedback when server-side auto-chunking doesn't kick in.

voyageai/object/contextualized_embeddings.py
jasonqiu95: we can assume the gateway always support enable auto chunking. It's a bigger problem if auto chunking is broken somehow on the server side and I'd surprised if things even work. IMO adding a warning to reflect the problem in the SDK is not a must do.

voyageai/object/contextualized_embeddings.py
jasonqiu95: I will make sure server returns a proper error response when results are somehow partial. I don't think partial results are useful. The error shouldn't happen any way, it's just defensive programming.

voyageai/object/contextualized_embeddings.py
jasonqiu95: Right now there is no caller that calls update multiple times. Even if there is, the original code seems broken/unclear anyway. It will accumulate the results but index will start afresh each batch. IMO we don't need to fix this as part of this PR.

fzoll (approved):
LGTM

fzoll (approved):
LGTM

fzowl (approved):
LGTM

PR #55, opened 2026-05-21
Fix async cancellation leaking aiohttp ClientSession

### LINKED ISSUE #52: Cancellations are not handled properly in python3.12+ : Async cancellation leaks aiohttp ClientSession in APIRequestor.arequest()
(Disclaimer: the investigation and repro were done with the help of Codex; I reviewed all of it)
### Summary

`APIRequestor.arequest()` appears to leak an `aiohttp.ClientSession` when the coroutine is cancelled after `AioHTTPSession.__aenter__()` but before cleanup runs.

The cleanup path currently runs on success or `except Exception`, but on Python 3.12+ `asyncio.CancelledError` is a `BaseException`, not an `Exception`, so `ctx.__aexit__()` is skipped and the session is left unclosed.
When the leftover resources is GC'd down the line, it causes an `Unclosed client session` error to be raised.

### Environment

- `voyageai==0.3.7`
- Python `3.12.11`
- `aiohttp==3.13.3`

### Relevant code

`voyageai/api_resources/api_requestor.py`

`APIRequestor.arequest()` currently does manual context management:

```py
ctx = AioHTTPSession()
session = await ctx.__aenter__()
...
except Exception:
    ...
    await ctx.__aexit__(None, None, None)
```
and AioHTTPSession.__aenter__() may create a fresh aiohttp.ClientSession().

### Minimal repro:
```
import asyncio
import gc
import io
import logging
import warnings

from voyageai.api_resources.api_requestor import APIRequestor

async def sleepy_arequest_raw(*args, **kwargs):
    await asyncio.sleep(3600)

async def main() -> None:
    loop = asyncio.get_running_loop()
    loop.set_debug(True)  # only to make the source traceback visible

    stream = io.StringIO()
    handler = logging.StreamHandler(stream)
    logger = logging.getLogger("asyncio")
    old_level = logger.level
    logger.setLevel(logging.ERROR)
    logger.addHandler(handler)

    requestor = APIRequestor(key="dummy")
    requestor.arequest_raw = sleepy_arequest_raw  # monkey-patch for repro

    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always", ResourceWarning)

        task = asyncio.create_task(requestor.arequest("get", "/fake"))
        await asyncio.sleep(0)  # let arequest() enter AioHTTPSession
        task.cancel()

        try:
            await task
        except asyncio.CancelledError:
            pass

        for _ in range(50):
            gc.collect()
            await asyncio.sleep(0)

    logger.removeHandler(handler)
    logger.setLevel(old_level)

    combined = stream.getvalue() + "\n" + "\n".join(str(w.message) for w in caught)
    print("HAS_UNCLOSED", "Unclosed client session" in combined)
    print(combined)


asyncio.run(main())
```

fzowl:
## Summary
- Replace `except Exception` with `try/finally` in `APIRequestor.arequest()` so that `asyncio.CancelledError` (a `BaseException` since Python 3.9) properly triggers session cleanup instead of leaking the `aiohttp.ClientSession`.

## What changed
The cleanup path in `arequest()` used manual `__aenter__`/`__aexit__` with an `except Exception` guard. Since `CancelledError` is a `BaseException`, cancelling the coroutine skipped `ctx.__aexit__()` entirely, leaving the session unclosed. Switching to `try/finally` ensures cleanup runs on success, exception, and cancellation alike.

## Test plan
- [x] 3 new unit tests covering cleanup on cancellation, exception, and success (no API key needed)
- [x] Full async integration test suite passes (20/20)

Fixes #52

### DIFF
--- a/tests/test_client_async.py
+++ b/tests/test_client_async.py
@@ -1,6 +1,10 @@
+import asyncio
+from unittest.mock import AsyncMock, patch
+
 import pytest
 import voyageai
 import voyageai.error as error
+from voyageai.api_resources.api_requestor import APIRequestor, AioHTTPSession
 from voyageai.chunking import default_chunk_fn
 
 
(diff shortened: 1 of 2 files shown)


fzowl:

PR #57, opened 2026-06-01
Fix async cancellation leaking aiohttp ClientSession

### LINKED ISSUE #52: Cancellations are not handled properly in python3.12+ : Async cancellation leaks aiohttp ClientSession in APIRequestor.arequest()

fzowl:
## Summary
- Replace `except Exception` with `try/finally` in `APIRequestor.arequest()` so that `asyncio.CancelledError` (a `BaseException` since Python 3.9) properly triggers session cleanup instead of leaking the `aiohttp.ClientSession`.

## What changed
The cleanup path in `arequest()` used manual `__aenter__`/`__aexit__` with an `except Exception` guard. Since `CancelledError` is a `BaseException`, cancelling the coroutine skipped `ctx.__aexit__()` entirely, leaving the session unclosed. Switching to `try/finally` ensures cleanup runs on success, exception, and cancellation alike.

## Test plan
- [x] 3 new unit tests covering cleanup on cancellation, exception, and success (no API key needed)
- [x] Full async integration test suite passes (20/20)

Fixes #52

(Reopened from #55 — moved to upstream branch for CI secret access)

### DIFF
--- a/tests/test_client_async.py
+++ b/tests/test_client_async.py
@@ -585,55 +585,70 @@ class TestAsyncClient:
 
 
 class TestAsyncCancellationCleanup:
-    """Verify that asyncio.CancelledError does not leak an aiohttp.ClientSession."""
+    """Verify that asyncio.CancelledError does not leak an aiohttp.ClientSession.
+
+    Each test injects a controlled session via ``voyageai.aiosession`` so that
+    no real ``aiohttp.ClientSession`` is created during the test.  The
+    cancellation test is the load-bearing one that distinguishes the buggy
+    (``except Exception``) path from the fixed (``except BaseException``) path;
+    the exception and success tests verify cleanup in the remaining branches for
+    completeness.
+    """
 
     @pytest.mark.asyncio
     async def test_arequest_cleans_up_session_on_cancellation(self):
         requestor = APIRequestor(key="dummy")
+        mock_session = AsyncMock()
+        token = voyageai.aiosession.set(mock_session)
 
         mock_exit = AsyncMock()
-        original_aenter = AioHTTPSession.__aenter__
-
-        async def patched_aenter(self_session):
-            session = await original_aenter(self_session)
-            return session
 
         async def slow_arequest_raw(*args, **kwargs):
             await asyncio.sleep(3600)
 
-        with (
-            patch.object(requestor, "arequest_raw", side_effect=slow_arequest_raw),
-            patch.object(AioHTTPSession, "__aexit__", mock_exit),
-        ):
-            task = asyncio.create_task(requestor.arequest("POST", "/test", params={}))
-            await asyncio.sleep(0.05)
-            task.cancel()
-            with pytest.raises(asyncio.CancelledError):
-                await task
+        try:
+            with (
+                patch.object(requestor, "arequest_raw", side_effect=slow_arequest_raw),
+                patch.object(AioHTTPSession, "__aexit__", mock_exit),
+            ):
+                task = asyncio.create_task(requestor.arequest("POST", "/test", params={}))
+                await asyncio.sleep(0.05)
+                task.cancel()
+                with pytest.raises(asyncio.CancelledError):
+                    await task
+        finally:
+            voyageai.aiosession.reset(token)
 
         mock_exit.assert_awaited_once()
 
     @pytest.mark.asyncio
     async def test_arequest_cleans_up_session_on_exception(self):
         requestor = APIRequestor(key="dummy")
+        mock_session = AsyncMock()
+        token = voyageai.aiosession.set(mock_session)
 
         mock_exit = AsyncMock()
 
         async def failing_arequest_raw(*args, **kwargs):
             raise ValueError("boom")
 
-        with (
-            patch.object(requestor, "arequest_raw", side_effect=failing_arequest_raw),
-            patch.object(AioHTTPSession, "__aexit__", mock_exit),
-        ):
-            with pytest.raises(ValueError, match="boom"):
-                await requestor.arequest("POST", "/test", params={})
+        try:
+            with (
+                patch.object(requestor, "arequest_raw", side_effect=failing_arequest_raw),
+                patch.object(AioHTTPSession, "__aexit__", mock_exit),
+            ):
+                with pytest.raises(ValueError, match="boom"):
+                    await requestor.arequest("POST", "/test", params={})
+        finally:
+            voyageai.aiosession.reset(token)
 
         mock_exit.assert_awaited_once()
 
     @pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


tests/test_client_async.py
```diff
@@ -580,3 +582,73 @@ def test_async_client_count_tokens(self):
...
+    @pytest.mark.asyncio
+    async def test_arequest_cleans_up_session_on_cancellation(self):
+        requestor = APIRequestor(key="dummy")
+
+        mock_exit = AsyncMock()
+        original_aenter = AioHTTPSession.__aenter__
+
+        async def patched_aenter(self_session):
+            session = await original_aenter(self_session)
+            return session
```
jasonqiu95: **Dead code.** `original_aenter` and `patched_aenter` are defined but never used — `patched_aenter` is never passed to a `patch.object(...)`, never assigned to `AioHTTPSession.__aenter__`, and never called (the `with` block below only patches `arequest_raw` and `__aexit__`). It also wouldn't change anything if it were wired up, since it just awaits the original and returns the same session. Suggest deleting these five lines to avoid implying `__aenter__` is being instrumented.

tests/test_client_async.py
```diff
@@ -580,3 +582,73 @@ def test_async_client_count_tokens(self):
...
+
+        async def slow_arequest_raw(*args, **kwargs):
+            await asyncio.sleep(3600)
+
+        with (
+            patch.object(requestor, "arequest_raw", side_effect=slow_arequest_raw),
+            patch.object(AioHTTPSession, "__aexit__", mock_exit),
```
jasonqiu95: **Replacing `AioHTTPSession.__aexit__` with an `AsyncMock` leaks a real `ClientSession`** (this applies to all three tests). Because `voyageai.aiosession.get()` is `None` by default, the real `__aenter__` runs `await aiohttp.ClientSession().__aenter__()` and sets `_should_close_session = True`. Mocking `__aexit__` then prevents the only code path that would close that session. Running these tests emits:

```
ResourceWarning: Unclosed client session <aiohttp.client.ClientSession object ...>
```

for each test (confirmed locally). Under `-W error::ResourceWarning` this would become a hard failure.

More importantly, this means the tests assert that `__aexit__` was **called** (`mock_exit.assert_awaited_once()`), not that the session was actually **closed** — which is the property the class docstring claims to verify. Consider, in setup, setting `voyageai.aiosession` to a controlled session (so `_should_close_session` stays `False` and no real socket is opened) and asserting against that, or spying on `__aexit__` with `wraps=` so the real close still runs. Either way add teardown that resets `voyageai.aiosession`.

tests/test_client_async.py
```diff
@@ -580,3 +582,73 @@ def test_async_client_count_tokens(self):
...
+        ):
+            task = asyncio.create_task(requestor.arequest("POST", "/test", params={}))
+            await asyncio.sleep(0.05)
+            task.cancel()
+            with pytest.raises(asyncio.CancelledError):
+                await task
+
+        mock_exit.assert_awaited_once()
+
+    @pytest.mark.asyncio
```
jasonqiu95, line 624: **`on_exception` and `on_success` don't actually guard the regression.** A `ValueError` and the success path are both handled by the *old* `except Exception:` code too, so I verified that these two tests **pass on both the buggy and fixed versions** — only `test_..._on_cancellation` distinguishes them (it fails on the old code). These two are still useful as general cleanup coverage, but they give no protection against re-introducing the `except Exception` bug. Worth a comment noting that the cancellation test is the load-bearing one, or strengthening these to also exercise a `BaseException`-but-not-`Exception` path.

jasonqiu95 (commented):
Reviewed the production change and the new tests, and ran the suite locally to confirm each claim.

**The core fix is correct and well-targeted.** `asyncio.CancelledError` subclasses `BaseException` (not `Exception`) on the supported Pythons (3.9–3.14), so the old `except Exception:` genuinely skipped cleanup on cancellation and leaked the `aiohttp.ClientSession`. The `try/finally` rewrite runs `result.release()` + `await ctx.__aexit__(...)` on success, exception, and cancellation alike, with the same cleanup order and unchanged return/exception contract. The single caller (`APIResource.acreate`) is unaffected. I verified empirically that `test_..._on_cancellation` **fails** on the old code and **passes** on the fix — it's a real regression guard. 👍

The notes below are all test-quality (non-blocking); the production change is good to merge.

jasonqiu95 (approved):

PR #47, opened 2026-01-27
Make ffmpeg package optional

fzowl:

PR #58, opened 2026-06-01
Make ffmpeg package optional

fzowl:
Reopened from #47 — moved to upstream branch for CI secret access.

---

**Previous discussion from #47:**

> **@fzliu** (review comment on `voyageai/video_utils.py:191`):
> Let's also add back `ensure ffmpeg is available on PATH` to this error message

> **@fzowl**:
> @fzliu Can you please check?

### DIFF
--- a/voyageai/__init__.py
+++ b/voyageai/__init__.py
@@ -36,19 +36,6 @@ from voyageai.embeddings_utils import (
 )
 from voyageai.version import VERSION
 
-
-def _is_ffmpeg_available() -> bool:
-    """Check if ffmpeg-python is installed."""
-    try:
-        import ffmpeg  # noqa: F401
-
-        return True
-    except ImportError:
-        return False
-
-
-HAS_VIDEO = _is_ffmpeg_available()
-
 if TYPE_CHECKING:
     import requests
     from aiohttp import ClientSession
--- a/voyageai/__init__.py
+++ b/voyageai/__init__.py
@@ -36,6 +36,19 @@ from voyageai.embeddings_utils import (
 )
 from voyageai.version import VERSION
 
+
+def _is_ffmpeg_available() -> bool:
+    """Check if ffmpeg-python is installed."""
+    try:
+        import ffmpeg  # noqa: F401
+
+        return True
+    except ImportError:
+        return False
+
+
+HAS_VIDEO = _is_ffmpeg_available()
+
 if TYPE_CHECKING:
     import requests
     from aiohttp import ClientSession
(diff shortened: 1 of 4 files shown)


jasonqiu95 (commented):
Code review (recall pass) on the optional-ffmpeg change. Diff is small and clean — the logic is correct. Two notes below: one actionable (the suggested install command isn't shell-safe), one minor question about `HAS_VIDEO`.

voyageai/video_utils.py
```diff
@@ -187,8 +187,9 @@ def _load_video_bytes(video: Union[str, PathLike[str], bytes, Video]) -> bytes:
 def _ensure_ffmpeg_available() -> None:
     if ffmpeg is None:
         raise ImportError(
-            "ffmpeg-python is required for video optimization. "
-            "Install `ffmpeg-python` and ensure `ffmpeg` is available on PATH."
+            "The 'ffmpeg-python' package is required for video optimization. "
+            "Install it with: pip install voyageai[video] "
```
jasonqiu95: **`pip install voyageai[video]` isn't shell-safe.** Unquoted square brackets are glob patterns in zsh (the default macOS shell since Catalina), so this exact command fails before pip ever runs:

```
$ zsh -c 'echo voyageai[video]'
zsh: no matches found: voyageai[video]
```

Since the README has no video/install docs, this error message is the only place users learn how to install the extra — so the command being copy-paste-broken on zsh is a real dead-end. Suggest quoting it so it works in both bash and zsh:

```suggestion
            "Install it with: pip install 'voyageai[video]' "
```

voyageai/__init__.py
```diff
@@ -36,6 +36,19 @@
...
+    """Check if ffmpeg-python is installed."""
+    try:
+        import ffmpeg  # noqa: F401
+
+        return True
+    except ImportError:
+        return False
+
+
+HAS_VIDEO = _is_ffmpeg_available()
```
jasonqiu95: `HAS_VIDEO` (and `_is_ffmpeg_available()`) aren't referenced anywhere in the package — nothing internally gates on the flag, since the real check is the lazy `_ensure_ffmpeg_available()` in `video_utils.py`. If it's intended as a public capability flag (e.g. `if voyageai.HAS_VIDEO:`), consider documenting it / adding a test so it doesn't silently break; if not, it's effectively dead code that runs an import probe at every `import voyageai`. Minor — just flagging the intent.

jasonqiu95 (approved):

PR #59, opened 2026-06-01
Local model (voyage-4-nano) support

fzowl:
Seamless integration of local model(s)

Reopened from #49 — moved to upstream branch for CI secret access.

## Tracked follow-ups (intentionally out of scope for this PR)

- #65 — Local models: batched encode instead of the serial inference lock (throughput). The current per-model lock is a correct, documented defensive fix; real parallelism needs request coalescing, which is a separate feature.
- #66 — Local models: expose a `trust_remote_code` opt-out. Deferred because the only supported local model (`voyage-4-nano`) requires it; revisit when a second/community model with a different trust boundary is added.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,17 +7,18 @@ on:
     branches: [ "main" ]
   workflow_call:
 
+env:
+  POETRY_HOME: "/opt/poetry"
+
 jobs:
-  test:
-    name: Tests
+  # Unit tests without local dependencies (tests import error handling)
+  unit-tests-no-local-deps:
+    name: Unit Tests (no local deps)
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
-    env:
-      VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
-      POETRY_HOME: "/opt/poetry"
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
@@ -38,7 +39,75 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y ffmpeg
         ffmpeg -version
-    - name: Install package
+    - name: Install package (without local extras)
       run: $POETRY_HOME/bin/poetry install
-    - name: Run tests
+    - name: Run unit tests (no API key, no local deps)
+      run: $POETRY_HOME/bin/poetry run pytest tests/test_client_local.py tests/test_client_local_async.py -v
+
+  # Local model tests with local dependencies but no API key
+  local-model-tests:
+    name: Local Model Tests
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install poetry
+      run: |
+        python3 -m venv $POETRY_HOME
+        $POETRY_HOME/bin/pip install poetry==1.8.4
+        $POETRY_HOME/bin/poetry --version
+    - name: Install ffmpeg 7
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y software-properties-common
+        sudo add-apt-repository -y ppa:ubuntuhandbook1/ffmpeg7
+        sudo apt-get update
+        sudo apt-get install -y ffmpeg
+        ffmpeg -version
+    - name: Install package with local extras
+      run: $POETRY_HOME/bin/poetry install --extras local
+    - name: Run local model tests (no API key)
+      run: $POETRY_HOME/bin/poetry run pytest tests/test_client_local.py tests/test_client_local_async.py -v
+
+  # Full integration tests with API key and all dependencies
+  integration-tests:
+    name: Integration Tests
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+    env:
+      VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install poetry
+      run: |
+        python3 -m venv $POETRY_HOME
+        $POETRY_HOME/bin/pip install poetry==1.8.4
+        $POETRY_HOME/bin/poetry --version
+    - name: Install ffmpeg 7
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y software-properties-common
+        sudo add-apt-repository -y ppa:ubuntuhandbook1/ffmpeg7
+        sudo apt-get update
+        sudo apt-get install -y ffmpeg
+        ffmpeg -version
+    - name: Install package with local extras
+      run: $POETRY_HOME/bin/poetry install --extras local
+    - name: Run all tests
       run: $POETRY_HOME/bin/poetry run pytest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,14 +45,13 @@ jobs:
       run: $POETRY_HOME/bin/poetry run pytest tests/test_client_local.py tests/test_client_local_async.py -v
 
   # Local model tests with local dependencies but no API key
-  # Requires Python 3.10+ because the HuggingFace model code uses 3.10+ syntax
   local-model-tests:
     name: Local Model Tests
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
@@ -81,14 +80,13 @@ jobs:
       run: $POETRY_HOME/bin/poetry run pytest tests/test_client_local.py tests/test_client_local_async.py -v
 
   # Full integration tests with API key and all dependencies
-  # Requires Python 3.10+ for local model deps (HuggingFace model uses 3.10+ syntax)
   integration-tests:
     name: Integration Tests
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
     env:
       VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
     steps:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -85,6 +85,11 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
+      # Run one Python version at a time: every job here hits the live API
+      # (incl. token-heavy voyage-multimodal-3.5), and running the whole matrix
+      # in parallel exceeds the project's per-minute token (TPM) rate limit,
+      # turning real assertions into spurious RateLimitError failures.
+      max-parallel: 1
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
     env:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,13 +45,14 @@ jobs:
       run: $POETRY_HOME/bin/poetry run pytest tests/test_client_local.py tests/test_client_local_async.py -v
 
   # Local model tests with local dependencies but no API key
+  # Requires Python 3.10+ because the HuggingFace model code uses 3.10+ syntax
   local-model-tests:
     name: Local Model Tests
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
@@ -72,19 +73,22 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y ffmpeg
         ffmpeg -version
+    - name: Install OpenBLAS (needed for scipy build on 3.13+)
+      run: sudo apt-get install -y libopenblas-dev
     - name: Install package with local extras
       run: $POETRY_HOME/bin/poetry install --extras local
     - name: Run local model tests (no API key)
       run: $POETRY_HOME/bin/poetry run pytest tests/test_client_local.py tests/test_client_local_async.py -v
 
   # Full integration tests with API key and all dependencies
+  # Requires Python 3.10+ for local model deps (HuggingFace model uses 3.10+ syntax)
   integration-tests:
     name: Integration Tests
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
     env:
       VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
     steps:
@@ -107,6 +111,8 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y ffmpeg
         ffmpeg -version
+    - name: Install OpenBLAS (needed for scipy build on 3.13+)
+      run: sudo apt-get install -y libopenblas-dev
     - name: Install package with local extras
       run: $POETRY_HOME/bin/poetry install --extras local
     - name: Run all tests
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,9 +21,9 @@ jobs:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install poetry
@@ -54,9 +54,9 @@ jobs:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install poetry
@@ -91,9 +91,9 @@ jobs:
       VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install poetry
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ pydantic = ">=1.10.8"
 tokenizers = ">=0.14.0"
 langchain-text-splitters = ">=0.3.8"
 ffmpeg-python = "*"
-sentence-transformers = {version = ">=3.0.0", optional = true, python = ">=3.10"}
+sentence-transformers = {version = ">=3.0.0,<5.0.0", optional = true, python = ">=3.10"}
 torch = {version = ">=2.9.0", optional = true, python = ">=3.10"}
 
 [tool.poetry.extras]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ tokenizers = ">=0.14.0"
 langchain-text-splitters = ">=0.3.8"
 ffmpeg-python = "*"
 sentence-transformers = {version = ">=3.0.0", optional = true}
-torch = {version = ">=2.0.0", optional = true}
+torch = {version = ">=2.9.0", optional = true}
 
 [tool.poetry.extras]
 local = ["sentence-transformers", "torch"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ pydantic = ">=1.10.8"
 tokenizers = ">=0.14.0"
 langchain-text-splitters = ">=0.3.8"
 ffmpeg-python = "*"
-sentence-transformers = {version = ">=3.0.0", optional = true}
+sentence-transformers = {version = ">=3.0.0", optional = true, python = ">=3.10"}
 torch = {version = ">=2.9.0", optional = true, python = ">=3.10"}
 
 [tool.poetry.extras]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "voyageai"
-version = "0.3.8"
+version = "0.3.9"
 description = ""
 authors = ["Yujie Qian <yujieq@voyageai.com>"]
 readme = "README.md"
@@ -20,6 +20,11 @@ pydantic = ">=1.10.8"
 tokenizers = ">=0.14.0"
 langchain-text-splitters = ">=0.3.8"
 ffmpeg-python = "*"
+sentence-transformers = {version = ">=3.0.0", optional = true}
+torch = {version = ">=2.0.0", optional = true}
+
+[tool.poetry.extras]
+local = ["sentence-transformers", "torch"]
 
 [tool.poetry.group.test.dependencies]
 pytest = "^7.4.2"
@@ -39,6 +44,12 @@ ignore = ["E501"]
 quote-style = "double"
 indent-style = "space"
 skip-magic-trailing-comma = false
+
+[tool.pytest.ini_options]
+markers = [
+    "integration: marks tests as integration tests (require external dependencies or API)",
+]
+asyncio_mode = "auto"
 
 [build-system]
 requires = ["poetry-core"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,10 +24,17 @@ ffmpeg-python = {version = "*", optional = true}
 sentence-transformers = {version = ">=5.0.0", optional = true, python = ">=3.10"}
 torch = {version = ">=2.9.0", optional = true, python = ">=3.10"}
 transformers = {version = ">=4.41.0,<5.0.0", optional = true, python = ">=3.10"}
+# scipy is pulled in transitively by sentence-transformers (via scikit-learn).
+# Pin a modern floor: scipy 1.6.1 declares `Requires-Python: >=3.7` with no
+# upper bound, so the resolver otherwise selects it for 3.13/3.14, where it has
+# no wheels and fails to build from source. 1.14.1 is the first release with
+# cp313 wheels; newer releases cover 3.14. (Only needed for the local extra,
+# which itself requires Python >=3.10.)
+scipy = {version = ">=1.14.1", optional = true, python = ">=3.10"}
 
 [tool.poetry.extras]
 video = ["ffmpeg-python"]
-local = ["sentence-transformers", "torch", "transformers"]
+local = ["sentence-transformers", "torch", "transformers", "scipy"]
 
 [tool.poetry.group.test.dependencies]
 pytest = "^7.4.2"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,11 +20,12 @@ pydantic = ">=1.10.8"
 tokenizers = ">=0.14.0"
 langchain-text-splitters = ">=0.3.8"
 ffmpeg-python = "*"
-sentence-transformers = {version = ">=3.0.0,<5.0.0", optional = true, python = ">=3.10"}
+sentence-transformers = {version = ">=5.0.0", optional = true, python = ">=3.10"}
 torch = {version = ">=2.9.0", optional = true, python = ">=3.10"}
+transformers = {version = ">=4.41.0,<5.0.0", optional = true, python = ">=3.10"}
 
 [tool.poetry.extras]
-local = ["sentence-transformers", "torch"]
+local = ["sentence-transformers", "torch", "transformers"]
 
 [tool.poetry.group.test.dependencies]
 pytest = "^7.4.2"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ tokenizers = ">=0.14.0"
 langchain-text-splitters = ">=0.3.8"
 ffmpeg-python = "*"
 sentence-transformers = {version = ">=3.0.0", optional = true}
-torch = {version = ">=2.9.0", optional = true}
+torch = {version = ">=2.9.0", optional = true, python = ">=3.10"}
 
 [tool.poetry.extras]
 local = ["sentence-transformers", "torch"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,9 +21,13 @@ pydantic = ">=2.7.4"
 tokenizers = ">=0.14.0"
 langchain-text-splitters = ">=0.3.8"
 ffmpeg-python = {version = "*", optional = true}
-sentence-transformers = {version = ">=5.0.0", optional = true, python = ">=3.10"}
+sentence-transformers = {version = ">=5.5.0", optional = true, python = ">=3.10"}
 torch = {version = ">=2.9.0", optional = true, python = ">=3.10"}
 transformers = {version = ">=4.41.0,<5.0.0", optional = true, python = ">=3.10"}
+# Floor is 5.5.0: the per-call processing_kwargs override that encode() relies on
+# to honor truncation=False was added there (PR #3753). On earlier 5.x the kwarg
+# is unknown, so truncation=False would be silently ignored and long inputs
+# truncated anyway.
 # scipy is pulled in transitively by sentence-transformers (via scikit-learn).
 # Pin a modern floor: scipy 1.6.1 declares `Requires-Python: >=3.7` with no
 # upper bound, so the resolver otherwise selects it for 3.13/3.14, where it has
--- a/tests/test_client_local.py
+++ b/tests/test_client_local.py
@@ -217,6 +217,45 @@ class TestLocalModelIntegration:
             norm = np.linalg.norm(result.embeddings[0])
             assert norm == pytest.approx(1.0, abs=1e-2), f"dim={dim} not unit-norm (norm={norm})"
 
+    def test_truncation_false_over_length_raises(self):
+        """truncation=False on over-length input raises InvalidRequestError, matching
+        the hosted API, instead of crashing/degrading inside the model."""
+        from voyageai import Client
+
+        client = Client()
+        long_text = "word " * 40000  # well beyond the 32768-token context limit
+
+        with pytest.raises(InvalidRequestError):
+            client.embed([long_text], model="voyage-4-nano", truncation=False)
+
+    def test_total_tokens_capped_when_truncated(self):
+        """With truncation on, total_tokens reports the processed (capped) count,
+        not the full pre-truncation length, matching the API."""
+        from voyageai import Client
+
+        client = Client()
+        long_text = "word " * 40000
+
+        result = client.embed([long_text], model="voyage-4-nano", truncation=True)
+        assert result.total_tokens <= 32768
+
+
+class TestLocalInputValidation:
+    """Empty / None input must raise InvalidRequestError like the hosted API.
+    These raise before any model load, so they need no local deps or network."""
+
+    def test_empty_list_raises(self):
+        from voyageai import Client
+
+        with pytest.raises(InvalidRequestError):
+            Client().embed([], model="voyage-4-nano")
+
+    def test_none_input_raises(self):
+        from voyageai import Client
+
+        with pytest.raises(InvalidRequestError):
+            Client().embed(None, model="voyage-4-nano")
+
 
 class TestKeylessApiGate:
     """A keyless client is allowed (for local models) but every API-backed method
@@ -258,3 +297,20 @@ class TestKeylessApiGate:
 
         with pytest.raises(voyageai.error.AuthenticationError):
             keyless_client.embed(["hello"], model="voyage-3")
+
+    def test_binary_key_file_falls_back_to_keyless(self, tmp_path, monkeypatch):
+        """A non-UTF-8 VOYAGE_API_KEY_PATH raises UnicodeDecodeError in
+        default_api_key(); construction must tolerate it and fall back to keyless
+        rather than crash."""
+        import voyageai
+        from voyageai import Client
+
+        key_file = tmp_path / "key.bin"
+        key_file.write_bytes(b"\xff\xfe\x00\x01")
+        monkeypatch.delenv("VOYAGE_API_KEY", raising=False)
+        monkeypatch.setattr(voyageai, "api_key", None, raising=False)
+        monkeypatch.setattr(voyageai, "api_key_path", None, raising=False)
+        monkeypatch.setenv("VOYAGE_API_KEY_PATH", str(key_file))
+
+        client = Client()  # must not raise UnicodeDecodeError
+        assert client.api_key is None
--- a/tests/test_client_local.py
+++ b/tests/test_client_local.py
@@ -13,6 +13,18 @@ try:
 except ImportError:
     REAL_DEPS_AVAILABLE = False
 
+# Check if local model can actually load (fails on Python <3.10 due to
+# HuggingFace model code using 3.10+ type union syntax)
+LOCAL_MODEL_WORKS = False
+if REAL_DEPS_AVAILABLE:
+    try:
+        from voyageai.local.sentence_transformer_backend import SentenceTransformerBackend
+
+        SentenceTransformerBackend("voyage-4-nano")
+        LOCAL_MODEL_WORKS = True
+    except (TypeError, Exception):
+        pass
+
 
 class TestLocalModelSupport:
     """Test local model detection and routing."""
@@ -43,9 +55,11 @@ class TestLocalModelIntegration:
 
     @pytest.fixture
     def check_deps(self):
-        """Skip if dependencies not installed."""
+        """Skip if dependencies not installed or model can't load."""
         if not REAL_DEPS_AVAILABLE:
             pytest.skip("sentence-transformers or torch not installed")
+        if not LOCAL_MODEL_WORKS:
+            pytest.skip("local model failed to load (requires Python 3.10+)")
 
     def test_seamless_local_embedding(self, check_deps):
         """Test that Client.embed() seamlessly uses local model."""
--- a/tests/test_client_local.py
+++ b/tests/test_client_local.py
@@ -1,20 +1,11 @@
 """Tests for local model support in Client."""
 
-import importlib.util
-
 import pytest
 from voyageai.error import InvalidRequestError
 
+from tests._local_test_utils import real_deps_available
 
-def _real_deps_available() -> bool:
-    """Check if sentence-transformers and torch can be found (without importing)."""
-    return (
-        importlib.util.find_spec("sentence_transformers") is not None
-        and importlib.util.find_spec("torch") is not None
-    )
-
-
-REAL_DEPS_AVAILABLE = _real_deps_available()
+REAL_DEPS_AVAILABLE = real_deps_available()
 
 
 @pytest.fixture(scope="session")
@@ -60,7 +51,7 @@ class TestLocalModelSupport:
         """Test HAS_LOCAL reflects dependency availability via independent check."""
         import voyageai
 
-        expected = _real_deps_available()
+        expected = real_deps_available()
         assert voyageai.HAS_LOCAL == expected
 
 
--- a/tests/test_client_local.py
+++ b/tests/test_client_local.py
@@ -1,0 +1,132 @@
+"""Tests for local model support in Client."""
+
+import pytest
+
+# ruff: noqa: F401
+
+# Check if real dependencies are available
+try:
+    import sentence_transformers
+    import torch
+
+    REAL_DEPS_AVAILABLE = True
+except ImportError:
+    REAL_DEPS_AVAILABLE = False
+
+
+class TestLocalModelSupport:
+    """Test local model detection and routing."""
+
+    @pytest.mark.skipif(REAL_DEPS_AVAILABLE, reason="Only run when deps not installed")
+    def test_import_error_when_deps_missing(self):
+        """Test helpful error message when sentence-transformers not installed."""
+        from voyageai.local import _ensure_local_deps
+
+        with pytest.raises(ImportError) as exc_info:
+            _ensure_local_deps()
+
+        assert "pip install voyageai[local]" in str(exc_info.value)
+
+    def test_has_local_constant(self):
+        """Test HAS_LOCAL constant reflects dependency availability."""
+        import voyageai
+
+        assert voyageai.HAS_LOCAL == REAL_DEPS_AVAILABLE
+
+
+@pytest.mark.integration
+class TestLocalModelIntegration:
+    """Integration tests for local models using the standard Client.
+
+    Run with: pytest -m integration
+    """
+
+    @pytest.fixture
+    def check_deps(self):
+        """Skip if dependencies not installed."""
+        if not REAL_DEPS_AVAILABLE:
+            pytest.skip("sentence-transformers or torch not installed")
+
+    def test_seamless_local_embedding(self, check_deps):
+        """Test that Client.embed() seamlessly uses local model."""
+        from voyageai import Client
+
+        # No API key needed for local models
+        client = Client()
+        result = client.embed(["Hello, world!"], model="voyage-4-nano", input_type="document")
+
+        assert len(result.embeddings) == 1
+        assert len(result.embeddings[0]) == 2048
+        assert result.total_tokens > 0
+
+    def test_all_dimensions(self, check_deps):
+        """Test all supported dimensions work."""
+        from voyageai import Client
+
+        client = Client()
+
+        for dim in [256, 512, 1024, 2048]:
+            result = client.embed(
+                ["Test text"], model="voyage-4-nano", input_type="document", output_dimension=dim
+            )
+            assert len(result.embeddings[0]) == dim, f"Expected {dim}, got {len(result.embeddings[0])}"
+
+    def test_float32_dtype(self, check_deps):
+        """Test float32 output data type (default)."""
+        from voyageai import Client
+
+        client = Client()
+
+        result = client.embed(["Test"], model="voyage-4-nano", input_type="document", output_dtype="float32")
+        assert isinstance(result.embeddings[0][0], float)
+
+    def test_query_vs_document_different(self, check_deps):
+        """Test query and document embeddings are different."""
+        from voyageai import Client
+
+        client = Client()
+
+        query_result = client.embed(["What is machine learning?"], model="voyage-4-nano", input_type="query")
+        doc_result = client.embed(["What is machine learning?"], model="voyage-4-nano", input_type="document")
+
+        # Embeddings should be different due to different prompts
+        assert query_result.embeddings[0] != doc_result.embeddings[0]
+
+    def test_batch_embedding(self, check_deps):
+        """Test batch embedding works."""
+        from voyageai import Client
+
+        client = Client()
+
+        texts = [
+            "First document",
+            "Second document",
+            "Third document",
+        ]
+        result = client.embed(texts, model="voyage-4-nano", input_type="document")
+
+        assert len(result.embeddings) == 3
+        for emb in result.embeddings:
+            assert len(emb) == 2048
+
+    def test_invalid_dimension_raises_error(self, check_deps):
+        """Test invalid dimension raises ValueError."""
+        from voyageai import Client
+
+        client = Client()
+
+        with pytest.raises(ValueError) as exc_info:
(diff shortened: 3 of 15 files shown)


voyageai/__init__.py
```diff
@@ -36,6 +36,20 @@
...
+    try:
+        import sentence_transformers  # noqa: F401
+        import torch  # noqa: F401
+
+        return True
+    except ImportError:
+        return False
+
+
+HAS_LOCAL = _is_local_available()
```
jasonqiu95: **[Correctness] `import voyageai` now eagerly imports torch + sentence-transformers, defeating the lazy-loading this PR promises.**

`HAS_LOCAL = _is_local_available()` runs `import torch` / `import sentence_transformers` at package load. Independently, `client.py` does `from voyageai.local.model_registry import ...` at top level, which executes `voyageai/local/__init__.py` — whose docstring says *"lazy imports to avoid loading torch/sentence-transformers for API-only users"* but then imports both at module body.

Net effect: every `import voyageai` for a user with `[local]` installed pays torch's multi-second import cost, even pure-API users. Consider deferring `HAS_LOCAL` to a lazy function/property and not eagerly importing in `local/__init__.py`.

voyageai/local/model_registry.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+# Supported local models
+SUPPORTED_MODELS: Dict[str, LocalModelConfig] = {
+    "voyage-4-nano": LocalModelConfig(
+        huggingface_id="voyageai/voyage-4-nano",
+        max_tokens=32768,
+        default_dimension=2048,
+        supported_dimensions=(2048, 1024, 512, 256),
+        supported_precisions=("float32", "int8", "uint8", "binary", "ubinary"),
```
jasonqiu95: **[Correctness] `output_dtype="float"` works against the API but raises locally.**

`supported_precisions` omits `"float"`, so `validate_precision("float")` (called at `sentence_transformer_backend.py:78`, *before* the `DTYPE_TO_PRECISION` map that does contain `"float"`) raises `ValueError: Invalid output_dtype 'float'`. The API path (`util._resolve_numpy_dtype`) accepts `"float"` and it's effectively the default dtype, so the same `client.embed(..., output_dtype="float")` call succeeds on an API model and fails on `voyage-4-nano`. Add `"float"` to `supported_precisions`.

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,111 @@
...
+        self.model = cache.get_or_load(cache_key, load_model)
+        self._tokenizer = self.model.tokenizer
+
+    def encode(
+        self,
+        texts: List[str],
+        input_type: Optional[str] = None,
+        output_dtype: Optional[str] = None,
+        output_dimension: Optional[int] = None,
+        truncation: bool = True,
```
jasonqiu95, line 119: **[Correctness] `truncation` is accepted but silently ignored, and `max_tokens` is never wired to the model.**

`truncation` never reaches `self.model.encode*()`, and `config.max_tokens=32768` is never applied to `model.max_seq_length`. So `truncation=False` yields silent model-side truncation instead of an error, and inputs are truncated at the checkpoint's default `max_seq_length` (which may be < 32768) rather than the advertised limit. The API path explicitly honors truncation (`_base.py` calls `tokenizer.no_truncation()`), so behavior diverges.

voyageai/_base.py
```diff
@@ -59,9 +59,18 @@ def __init__(
         timeout: Optional[float] = None,
         base_url: Optional[str] = None,
     ) -> None:
-        self.api_key = api_key or default_api_key()
+        # API key is optional - allow None for local-only usage
+        try:
+            self.api_key = api_key or default_api_key()
+        except voyageai.error.AuthenticationError:
```
jasonqiu95: **[Correctness/Altitude] api-key-optional + local routing is a per-method bandaid added only to `embed()`.**

`__init__` now swallows `AuthenticationError` → `self.api_key = None`, but only `embed()` (`client.py:120`, `client_async.py:141`) re-checks the key and routes `model in LOCAL_MODELS`. `contextualized_embed`, `rerank`, and `multimodal_embed` do neither — calling them with `"voyage-4-nano"` routes a local model name to the API path, and their no-key error surfaces deep inside `APIRequestor` instead of the friendly message `embed` gives. Consider a single shared dispatch/classifier so routing + key-checking can't drift across the 4 methods.

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,111 @@
...
+
+        Args:
+            texts: List of texts to count tokens for.
+
+        Returns:
+            Total token count.
+        """
+        total = 0
+        for text in texts:
+            encoded = self._tokenizer.encode(text, add_special_tokens=True)
```
jasonqiu95: **[Correctness] `count_tokens` undercounts for query/document inputs.**

It tokenizes the raw text only, but `encode_query`/`encode_document` prepend an instruction prefix that the model actually tokenizes and processes. The reported `total_tokens` therefore omits the prompt-prefix tokens and diverges from server-side usage accounting (undercount = prefix-token-count × num_texts).

voyageai/local/model_registry.py
```diff
@@ -0,0 +1,154 @@
...
+        """Get a cached model or load it if not cached.
+
+        Args:
+            key: Cache key.
+            loader: Callable to load the model if not cached.
+
+        Returns:
+            The cached or newly loaded model.
+        """
+        model = self.get(key)
```
jasonqiu95, line 202: **[Correctness] TOCTOU race in `get_or_load`, despite the "Thread-safe singleton" docstring.**

`get()` → `loader()` → `set()` runs without holding `_cache_lock` across the whole operation, so two concurrent first-time callers each execute the expensive `loader()` (loading the model twice). The async path (`asyncio.to_thread`) and the unlocked per-client `self._local_backends` dict make concurrent first calls realistic. Hold the lock for the full check-load-store, or use a per-key lock.

voyageai/local/model_registry.py
```diff
@@ -0,0 +1,154 @@
...
+
+# Supported local models
+SUPPORTED_MODELS: Dict[str, LocalModelConfig] = {
+    "voyage-4-nano": LocalModelConfig(
+        huggingface_id="voyageai/voyage-4-nano",
+        max_tokens=32768,
+        default_dimension=2048,
+        supported_dimensions=(2048, 1024, 512, 256),
+        supported_precisions=("float32", "int8", "uint8", "binary", "ubinary"),
+        trust_remote_code=True,
```
jasonqiu95, line 85: **[Security] `trust_remote_code=True` by default with no opt-out.**

This executes arbitrary code from the HuggingFace repo at model load. Defensible for a first-party `voyageai/` repo, but there's no way for a user to disable it. Consider exposing it as a constructor/config option (default could stay True) so security-sensitive deployments can turn it off.

pyproject.toml
```diff
@@ -20,6 +20,12 @@ pydantic = ">=1.10.8"
 tokenizers = ">=0.14.0"
 langchain-text-splitters = ">=0.3.8"
 ffmpeg-python = "*"
+sentence-transformers = {version = ">=5.0.0", optional = true, python = ">=3.10"}
```
jasonqiu95: **[Packaging] `[local]` extras are gated `python = ">=3.10"` while the package supports 3.9.**

On Python 3.9, `pip install voyageai[local]` silently installs *none* of sentence-transformers/torch/transformers (the markers exclude them), yet `model in LOCAL_MODELS` still routes to the local path and fails later with a confusing ImportError. Either document/raise a clear 3.10+ requirement for local models up front, or surface it at the routing boundary.

tests/test_client_local.py
```diff
@@ -0,0 +1,154 @@
...
+    REAL_DEPS_AVAILABLE = False
+
+# Check if local model can actually load (fails on Python <3.10 due to
+# HuggingFace model code using 3.10+ type union syntax)
+LOCAL_MODEL_WORKS = False
+if REAL_DEPS_AVAILABLE:
+    try:
+        from voyageai.local.sentence_transformer_backend import SentenceTransformerBackend
+
+        SentenceTransformerBackend("voyage-4-nano")
```
jasonqiu95, line 23: **[Tests] Model is loaded at import/collection time and all errors are swallowed.**

`SentenceTransformerBackend("voyage-4-nano")` runs at module top level, so pytest downloads/loads the model during *collection* — before marker filtering, even for `-m "not integration"` runs. The wrapping `except (TypeError, Exception): pass` (the `TypeError` is redundant — it's a subclass of `Exception`) hides every real failure (OOM, network, version mismatch), leaving `LOCAL_MODEL_WORKS=False` with no signal. Move this probe into a fixture/helper and narrow the except.

tests/test_client_local.py
```diff
@@ -0,0 +1,154 @@
...
+        with pytest.raises(ImportError) as exc_info:
+            _ensure_local_deps()
+
+        assert "pip install voyageai[local]" in str(exc_info.value)
+
+    def test_has_local_constant(self):
+        """Test HAS_LOCAL constant reflects dependency availability."""
+        import voyageai
+
+        assert voyageai.HAS_LOCAL == REAL_DEPS_AVAILABLE
```
jasonqiu95: **[Tests] `test_has_local_constant` is tautological.**

`voyageai.HAS_LOCAL` (via `_is_local_available`) and `REAL_DEPS_AVAILABLE` are computed by the *identical* `try: import sentence_transformers, torch`. The assertion can never fail and tests nothing. To make it meaningful, assert against an independent signal (e.g. `importlib.util.find_spec`).

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,111 @@
...
+        if input_type == "query":
+            embeddings = self.model.encode_query(texts, **encode_kwargs)
+        elif input_type == "document":
+            embeddings = self.model.encode_document(texts, **encode_kwargs)
+        else:
+            embeddings = self.model.encode(texts, **encode_kwargs)
+
+        return embeddings
+
+    def count_tokens(self, texts: List[str]) -> int:
```
jasonqiu95: **[Efficiency] Double tokenization per `embed` call.**

`encode()` already tokenizes internally, then `count_tokens()` re-tokenizes every text again — and does so in a Python per-text loop rather than a single `encode_batch(...)`. On large batches this doubles tokenization work. Consider batching and/or deriving the count from the encode pass.

voyageai/client_async.py
```diff
@@ -57,6 +63,62 @@ def _make_retry_controller(self) -> AsyncRetrying:
...
 
+    def _get_local_backend(self, model: str) -> "SentenceTransformerBackend":
+        """Get or create a local backend for the given model."""
+        if model not in self._local_backends:
+            from voyageai.local.sentence_transformer_backend import SentenceTransformerBackend
+
+            self._local_backends[model] = SentenceTransformerBackend(model)
+        return self._local_backends[model]
+
+    def _embed_local_sync(
```
jasonqiu95: **[Reuse] `_embed_local_sync`/`_embed_local` and `_get_local_backend` are duplicated verbatim between `client.py` and `client_async.py`.**

The sync bodies are byte-identical. Bug fixes must be applied twice and risk drift. Also note three overlapping dep-availability checks (`__init__._is_local_available`/`HAS_LOCAL`, `local.__init__._ensure_local_deps`, and the test-file probes). Consider a shared helper module.

jasonqiu95 (commented):
Automated code review (recall-oriented). 14 findings below, ranked by severity. I traced each against the branch; a few candidates were dropped as refuted (no-key `rerank`/`contextualized_embed`/`multimodal_embed` re-raise a clean `AuthenticationError` via `APIRequestor`, not a `Bearer None` crash; `.tolist()` is safe since ST v5 returns ndarrays; binary returning `dim/8` ints is API-parity-correct; and the `integration-tests` CI job still runs the full `pytest`).

**Correctness:** 1 eager torch import, 2 `output_dtype="float"` local/API divergence, 3 dropped `truncation`/`max_seq_length`, 4 local routing only in `embed()`, 5 token undercount, 6 cache race. **Security/packaging:** 7 `trust_remote_code=True`, 8 `[local]` gated to 3.10 on a 3.9 package. **Tests:** 9 collection-time model load, 10 tautological assert. **Cleanup:** 11-14.

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,111 @@
...
+            truncation: Whether to truncate texts exceeding max tokens.
+
+        Returns:
+            Numpy array of embeddings.
+        """
+        # Validate and get dimension
+        dimension = self.config.validate_dimension(output_dimension)
+
+        # Validate and map precision
+        self.config.validate_precision(output_dtype)
```
jasonqiu95: **[Simplification] `validate_precision`'s return value is discarded here.**

It's called purely for its raise-on-invalid side effect, then `DTYPE_TO_PRECISION.get(...)` recomputes the mapping. Either use the returned value or make the validation explicit. Relatedly, `DTYPE_TO_PRECISION` is near-identity (only `"float"→"float32"` differs from key==value), so it could collapse to a small special-case.

voyageai/client.py
```diff
@@ -57,6 +62,42 @@ def _make_retry_controller(self) -> Retrying:
             ),
         )
 
+    def _get_local_backend(self, model: str) -> "SentenceTransformerBackend":
```
jasonqiu95: **[Simplification] Two caches for one resource.**

The per-client `self._local_backends` dict and the global `ModelCache` singleton (keyed `model:device`) both cache the backend/model. The singleton already dedupes the heavy model process-wide, so the per-client dict mostly adds a second, unsynchronized layer. Consider keeping one (the singleton) to avoid split-brain state and the unlocked dict race.

jasonqiu95 (changes requested):

voyageai/_base.py
fzowl: Good catch on the inconsistency. For this PR, local model support is intentionally scoped to `embed()` only — voyage-4-nano is an embedding model, and there are no local equivalents for rerank, multimodal_embed, or contextualized_embed yet. Adding routing stubs there would be premature abstraction. That said, you're right that the error UX when calling those methods without an API key could be clearer — I'll note that for a follow-up.

voyageai/local/model_registry.py
fzowl: This is intentional for the current scope. The only supported local model (voyage-4-nano) requires `trust_remote_code=True` to function — its architecture uses custom modeling code hosted on HuggingFace. Exposing it as a configurable option would add complexity without practical benefit since disabling it would break the only supported model. We'll revisit this when/if we add support for third-party or community models where the trust boundary is different.

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,161 @@
...
+        precision = self.config.validate_precision(output_dtype)
+        if precision:
+            precision = DTYPE_TO_PRECISION[precision]
+
+        # Build encode kwargs
+        encode_kwargs = {}
+        if dimension != self.config.default_dimension:
+            encode_kwargs["truncate_dim"] = dimension
+        if precision:
+            encode_kwargs["precision"] = precision
```
jasonqiu95, line 180: **Blocker: `int8`/`uint8` embeddings are broken.** `encode(precision="int8")` is called with no fixed calibration `ranges`, so sentence-transformers derives the quantization range from *the current batch only*. A single-text call has `min==max` per dim → every value collapses to `-128`; the same text in different batches quantizes to different codes (non-deterministic, and incomparable to API/DB int8 vectors which use a fixed range).

Proposed fix — supply fixed calibration ranges (store them in `LocalModelConfig` so they match the API), e.g.:
```python
if precision in ("int8", "uint8"):
    encode_kwargs["precision"] = precision
    encode_kwargs["ranges"] = self.config.calibration_ranges  # fixed [2, dim] min/max
elif precision:
    encode_kwargs["precision"] = precision
```
If fixed ranges aren't available yet, reject `int8`/`uint8` locally rather than returning silently-wrong vectors.

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # Validate and map precision — use returned value directly
+        precision = self.config.validate_precision(output_dtype)
+        if precision:
+            precision = DTYPE_TO_PRECISION[precision]
+
+        # Build encode kwargs
+        encode_kwargs = {}
+        if dimension != self.config.default_dimension:
+            encode_kwargs["truncate_dim"] = dimension
```
jasonqiu95, line 178: **Blocker: truncated (MRL) embeddings aren't re-normalized.** `truncate_dim` slices the already-normalized 2048-d vector, so sub-2048 outputs are no longer unit-norm (measured on the cached model: 256-d → norm 0.327, 512-d → 0.476, 1024-d → 0.698; full-2048 = 1.0). The Voyage API returns unit-norm Matryoshka embeddings at every dimension, so cosine/dot scores diverge between local and API.

Proposed fix — re-normalize after truncation whenever a non-default dimension is requested:
```python
if dimension != self.config.default_dimension:
    encode_kwargs["truncate_dim"] = dimension
    encode_kwargs["normalize_embeddings"] = True
```

tests/test_client_local_async.py
```diff
@@ -0,0 +1,120 @@
...
+# Check if local model can actually load (fails on Python <3.10 due to
+# HuggingFace model code using 3.10+ type union syntax)
+LOCAL_MODEL_WORKS = False
+if REAL_DEPS_AVAILABLE:
+    try:
+        from voyageai.local.sentence_transformer_backend import SentenceTransformerBackend
+
+        SentenceTransformerBackend("voyage-4-nano")
+        LOCAL_MODEL_WORKS = True
+    except (TypeError, Exception):
```
jasonqiu95: **Blocker: this masks a broken backend in CI.** `except (TypeError, Exception)` catches *everything* (the `TypeError` is redundant), so any real failure loading the model — download error, OOM, runtime error — is swallowed, `LOCAL_MODEL_WORKS` stays `False`, and every async integration test silently skips with the misleading "requires Python 3.10+" reason. A genuinely broken local model would show as green CI.

Proposed fix — catch only the documented 3.10-syntax case and let everything else fail loudly:
```python
    except TypeError:
        pass  # 3.10+ union syntax in remote model code; other errors should surface
```

voyageai/_base.py
jasonqiu95: **Blocker: keyless clients now defer `AuthenticationError` for non-`embed` methods.** Swallowing the error here (and setting `api_key=None`) is paired with a re-check that was added *only* to `embed`. Previously `Client()` failed at construction with a clear message; now a keyless `rerank` / `multimodal_embed` / `contextualized_embed` call runs input validation/chunking (wasted work) and then raises the same error late and deep inside `api_requestor.py` via the retry loop.

Proposed fix — add one shared gate instead of per-method special-casing, e.g. a `_require_api_key()` helper called at the top of every API-backed method (and skipped on the local-model branch):
```python
def _require_api_key(self):
    if not self.api_key:
        raise error.AuthenticationError(
            "An API key is required for API-based models. Set VOYAGE_API_KEY "
            "or pass api_key=... to the client."
        )
```
Then call `self._require_api_key()` in `rerank`/`multimodal_embed`/`contextualized_embed` (and reuse it in `embed`).

voyageai/local/helpers.py
```diff
@@ -0,0 +1,69 @@
...
+    embeddings_array, total_tokens = backend.encode(
+        texts=texts,
+        input_type=input_type,
+        output_dtype=output_dtype,
+        output_dimension=output_dimension,
+        truncation=truncation,
+    )
+
+    result = EmbeddingsObject()
+    result.embeddings = embeddings_array.tolist()
```
jasonqiu95, line 87: **Blocker: bare-string input returns a flat 1-D list locally.** The API path tolerates a bare string (`vo.embed("text", ...)` → `embeddings == [[...]]`; exercised in `tests/test_client.py`). Locally, `encode_query("text")` returns a 1-D array, so `.tolist()` yields `[float, float, ...]` and `result.embeddings[0]` is a `float`, not a vector.

Proposed fix — normalize to a list before encoding (e.g. at the top of `embed_local`):
```python
if isinstance(texts, str):
    texts = [texts]
```

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,161 @@
...
+            encode_kwargs["processing_kwargs"] = {
+                "text": {"truncation": False},
+            }
+
+        # Route based on input_type using prompt_name
+        if input_type == "query":
+            embeddings = self.model.encode_query(texts, **encode_kwargs)
+        elif input_type == "document":
+            embeddings = self.model.encode_document(texts, **encode_kwargs)
+        else:
```
jasonqiu95: **Blocker: invalid `input_type` is silently ignored.** The API raises `InvalidRequestError` for a bad `input_type` (e.g. `"doc"`; see `tests/test_client.py`). Here, anything other than `"query"`/`"document"` falls into this `else` and runs `encode()` with no prompt, so a typo like `"doc"` or `"Query"` silently degrades embedding quality with no error.

Proposed fix — validate up front and only treat `None` as the no-prompt case:
```python
if input_type is not None and input_type not in ("query", "document"):
    raise ValueError(
        f"Invalid input_type {input_type!r}. Use 'query', 'document', or None."
    )
```

jasonqiu95 (commented):
Review focused on the merge-blockers: local embeddings must match the hosted API (otherwise vectors from the two paths are incomparable), plus the CI gap that would hide such regressions. Each inline comment includes a proposed fix.

jasonqiu95 (changes requested):

jasonqiu95 (commented):
## Automated code review — local model (voyage-4-nano) support

Reviewed the full diff at extra-high effort. Verified findings empirically in a venv against the PR branch (sentence-transformers 5.6): the local unit tests pass, lazy imports work (torch is not imported by `import voyageai`), `HAS_LOCAL` resolves correctly, and the keyless client routes as intended.

**Two confirmed correctness bugs** (reproduced):
1. `ModelCache` double-checked-locking race — publishes the singleton before `_cache_lock` exists; crashes under the concurrent async path.
2. A stale `VOYAGE_API_KEY_PATH` raises `FileNotFoundError` that the keyless `except AuthenticationError` doesn't catch, breaking local-only client construction.

Plus a likely normalization inconsistency, a version-pin risk, and several simplification/altitude/reuse items. Details inline.

voyageai/local/model_registry.py
```diff
@@ -0,0 +1,164 @@
...
+    _lock: threading.Lock = threading.Lock()
+
+    def __new__(cls) -> "ModelCache":
+        if cls._instance is None:
+            with cls._lock:
+                if cls._instance is None:
+                    cls._instance = super().__new__(cls)
+                    cls._instance._cache: Dict[str, Any] = {}
+                    cls._instance._cache_lock = threading.Lock()
+        return cls._instance
```
jasonqiu95, line 140: **Bug (race): the double-checked-locking singleton publishes `cls._instance` before `_cache`/`_cache_lock` are initialized.**

Root cause: line 115 assigns `cls._instance = super().__new__(cls)` and only then (116–117) sets `_cache` and `_cache_lock`. The *outer* `if cls._instance is None` check on line 112 runs **without the lock**. A second thread that sees the freshly published-but-not-yet-initialized instance returns it and immediately calls `get()`/`set()`/`get_or_load()`, which dereference `self._cache_lock` → `AttributeError: 'ModelCache' object has no attribute '_cache_lock'`.

This is exactly the concurrency the async path exercises: `AsyncClient.embed` → `asyncio.to_thread(embed_local, …)` → `ModelCache()` from multiple worker threads on the first calls. I reproduced the crash with 8 threads and a small window between lines 115 and 117.

Fix — fully initialize a local instance, then publish it last:

```python
def __new__(cls) -> "ModelCache":
    if cls._instance is None:
        with cls._lock:
            if cls._instance is None:
                instance = super().__new__(cls)
                instance._cache = {}
                instance._cache_lock = threading.Lock()
                cls._instance = instance  # publish only after fully initialized
    return cls._instance
```
(The annotated form `cls._instance._cache: Dict[str, Any] = {}` does assign at runtime — the bug is purely the publish-before-init ordering.)

voyageai/_base.py
```diff
@@ -59,16 +59,40 @@ def __init__(
...
+            # No API key available - that's OK for local models
+            self.api_key = None
```
jasonqiu95, line 72: **Bug: only `AuthenticationError` is caught, but `default_api_key()` raises `FileNotFoundError`/`OSError` when `VOYAGE_API_KEY_PATH` points at an unreadable file.**

Root cause: `voyageai/util.py:default_api_key()` does `open(api_key_path, "rt")` and only raises `AuthenticationError` when *no* key/path is configured. A set-but-stale `VOYAGE_API_KEY_PATH` makes `open()` raise `FileNotFoundError`, which escapes this `except` and breaks construction of a keyless local-only client — the very case this block is meant to allow.

Reproduced: with `VOYAGE_API_KEY` unset and `VOYAGE_API_KEY_PATH=/nonexistent/key.txt`, `Client()` raises `FileNotFoundError: [Errno 2] No such file or directory` instead of falling back to `api_key=None`.

Fix — also treat an unreadable key file as "no key":

```python
try:
    self.api_key = api_key or default_api_key()
except (voyageai.error.AuthenticationError, OSError):
    # No usable API key — fine for local models
    self.api_key = None
```

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,205 @@
...
+        # Build encode kwargs
+        encode_kwargs = {}
+        if dimension != self.config.default_dimension:
+            # Truncating an already unit-normalized vector (Matryoshka / MRL)
+            # breaks unit norm, so re-normalize to match the API, which returns
+            # unit-norm embeddings at every dimension.
+            encode_kwargs["truncate_dim"] = dimension
+            encode_kwargs["normalize_embeddings"] = True
+        if precision:
+            encode_kwargs["precision"] = precision
```
jasonqiu95: **Likely bug: the default-dimension path never sets `normalize_embeddings=True`, so full-dim output may not be unit-norm while truncated dims are.**

Root cause: `normalize_embeddings=True` is added only inside the `if dimension != self.config.default_dimension:` branch (line 154). For the default 2048-dim (and `output_dimension=None`) request, no `normalize_embeddings` is passed, and `SentenceTransformer.encode`/`encode_query`/`encode_document` default it to **`False`** (confirmed on sentence-transformers 5.6). Unless the model's own pipeline contains a Normalize layer, full-dim vectors come back un-normalized while truncated dims are normalized — inconsistent with each other and with the hosted API, which returns unit-norm at every dimension.

`test_truncated_dimensions_are_unit_norm` does cover dim 2048, so this will surface as a test failure if the model doesn't self-normalize — but it relies on implicit model behavior rather than the explicit guarantee the truncated path makes.

Fix — always normalize:

```python
encode_kwargs["normalize_embeddings"] = True
if dimension != self.config.default_dimension:
    encode_kwargs["truncate_dim"] = dimension
```

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,205 @@
...
+
+        # Wire truncation through to the model via processing_kwargs
+        if not truncation:
+            encode_kwargs["processing_kwargs"] = {
+                "text": {"truncation": False},
+            }
```
jasonqiu95: **Possible breakage on older `sentence-transformers` 5.x: `processing_kwargs` is only honored in recent releases.**

Root cause: `pyproject.toml` pins `sentence-transformers >= 5.0.0`, but the `processing_kwargs={"text": {"truncation": False}}` keyword and its allowlist in `encode()` were added after 5.0. On 5.6 it works (verified), but on an early 5.x an unknown kwarg either raises or is silently dropped — in which case `truncation=False` is silently ignored and long inputs get truncated anyway (silently wrong, not a crash).

Suggest either bumping the floor to the first release that supports `processing_kwargs`, or asserting the behavior in a test so the silent-truncation case can't ship undetected.

voyageai/__init__.py
```diff
@@ -36,6 +36,37 @@
...
+class _Module:
+    HAS_LOCAL = _LazyLocalFlag()
+
+
+_module_proxy = _Module()
+
+
+def __getattr__(name: str):
+    if name == "HAS_LOCAL":
+        return _module_proxy.HAS_LOCAL
```
jasonqiu95: **Over-engineered: four layers to expose one lazy boolean.**

`_is_local_available` (a wrapper around `voyageai.local._is_local_available`) + `_LazyLocalFlag` descriptor + `_Module` proxy class + `_module_proxy` instance + module-level `__getattr__` all exist just so `voyageai.HAS_LOCAL` lazily calls the availability check. A module `__getattr__` already provides exactly this laziness on its own — the descriptor/proxy indirection adds nothing.

Equivalent, ~6 lines:

```python
def __getattr__(name: str):
    if name == "HAS_LOCAL":
        from voyageai.local import _is_local_available
        return _is_local_available()
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```
Cost of keeping it: every reader has to trace the descriptor→proxy→getattr chain to confirm it does what the one-liner does.

voyageai/_base.py
```diff
@@ -59,16 +59,40 @@ def __init__(
...
+        would do its input validation/chunking and then fail late and deep inside
+        the request layer with a less actionable error.
+        """
+        if not self.api_key:
+            raise voyageai.error.AuthenticationError(
+                "An API key is required for API-based models. "
+                "Set your API key via the VOYAGE_API_KEY environment variable or "
+                "pass it to the client (api_key=...)."
+            )
+
```
jasonqiu95, line 100: **Altitude: the keyless design relies on every API method remembering to call `_require_api_key()` — a guard that's easy to forget on the next method added.**

The four current API methods are gated, so this isn't a present-day bug, but the pattern is fragile: `count_usage()` and `tokenize()` already don't call it (correctly, since they're keyless), so there's no consistent rule a future contributor can rely on. Consider making `api_key` a property that raises `AuthenticationError` on *access* when unset (local paths never read it), so the guarantee lives in one place instead of being re-asserted per method.

tests/test_client_local_async.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def _real_deps_available() -> bool:
+    """Check if sentence-transformers and torch can be found (without importing)."""
+    return (
+        importlib.util.find_spec("sentence_transformers") is not None
+        and importlib.util.find_spec("torch") is not None
+    )
+
+
```
jasonqiu95: **Reuse: `_real_deps_available()` is copy-pasted here and in `tests/test_client_local.py` (lines 9–16), and it duplicates `voyageai.local._is_local_available()`.**

Three near-identical copies of the same dep-probe. Extract one helper (e.g. a small `tests/conftest.py` fixture/util, or just call `voyageai.local._is_local_available`) so the Python-version / dep logic lives in a single place. As-is, any change to what counts as "deps available" has to be made in three spots.

voyageai/local/helpers.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Args:
+        model: Model name (must be in LOCAL_MODELS).
+
+    Returns:
+        SentenceTransformerBackend instance (cached via ModelCache singleton).
+    """
+    from voyageai.local.sentence_transformer_backend import SentenceTransformerBackend
+
+    return SentenceTransformerBackend(model)
```
jasonqiu95, line 29: **Misleading docstring + per-call wrapper rebuild.** The docstring says the returned `SentenceTransformerBackend` is "cached via ModelCache singleton," but only the underlying `SentenceTransformer` *model* is cached. `get_local_backend()` constructs a **new** `SentenceTransformerBackend` on every `embed_local` call, which re-runs `_ensure_local_deps()`, `get_model_config()`, the `ModelCache()` lookup, and `self._tokenizer = self.model.tokenizer` each time.

The per-call cost is modest (the heavy model load is cached), but the docstring implies a backend-level cache that doesn't exist. Either cache the backend object too (keyed like the model) or drop this one-line wrapper and call `SentenceTransformerBackend(model)` directly from `embed_local`, and fix the docstring to describe the actual (model-only) caching.

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,205 @@
...
+            encoded = self._tokenizer.encode(full_text, add_special_tokens=True)
+            total += len(encoded)
+        return total
+
+    def count_tokens(self, texts: List[str], input_type: Optional[str] = None) -> int:
+        """Count total tokens across all texts.
+
+        Args:
+            texts: List of texts to count tokens for.
+            input_type: "query", "document", or None.
```
jasonqiu95: **Simplification + efficiency.** Two points on token counting:

1. `count_tokens()` is a pure passthrough to `_count_tokens_with_prefix()` — collapse the two (make the prefixed one public, or have one delegate without a second public name).
2. `encode()` (line 173) calls `_count_tokens_with_prefix()` *after* the model has already tokenized the same texts during `encode`/`encode_query`/`encode_document`, and the counter re-tokenizes `f"{prompt}{text}"` per text in a Python loop. For a batch of N this re-tokenizes the (constant) instruction prompt N times. If exact parity with the model's own tokenization isn't required, prefer a single batched encode.

voyageai/local/model_registry.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        with self._cache_lock:
+            self._cache[key] = model
+
+    def clear(self) -> None:
+        """Clear all cached models."""
+        with self._cache_lock:
+            self._cache.clear()
+
+    def get_or_load(self, key: str, loader: callable) -> Any:
```
jasonqiu95: **Type hint: `loader: callable` should be `typing.Callable`.** `callable` is the builtin predicate function, not a type — type checkers (mypy/pyright) flag it and IDEs give no signature help. It happens to not error at runtime only because annotations aren't evaluated here.

```suggestion
    def get_or_load(self, key: str, loader: Callable[[], Any]) -> Any:
```
(add `Callable` to the existing `from typing import` line.)

voyageai/local/model_registry.py
```diff
@@ -0,0 +1,164 @@
...
+SUPPORTED_MODELS: Dict[str, LocalModelConfig] = {
+    "voyage-4-nano": LocalModelConfig(
+        huggingface_id="voyageai/voyage-4-nano",
+        max_tokens=32768,
+        default_dimension=2048,
+        supported_dimensions=(2048, 1024, 512, 256),
+        # int8/uint8 omitted: they require fixed calibration ranges to match the
+        # hosted API, which we don't ship locally yet (see encode() in
+        # sentence_transformer_backend.py).
+        supported_precisions=("float32", "float", "binary", "ubinary"),
```
jasonqiu95, line 84: **Simplification: int8/uint8 rejection is encoded in two places, and `supported_precisions` carries a redundant pair.**

int8/uint8 are excluded here *and* re-rejected with `NotImplementedError` via `_UNSUPPORTED_QUANTIZED_DTYPES` in `sentence_transformer_backend.encode()`. With the explicit `NotImplementedError` gate, listing the *supported* set is the single source of truth; the dual gate means a future edit to one can silently disagree with the other. Also `("float32", "float", …)` lists two names that map to the same precision in `DTYPE_TO_PRECISION` — fine, but worth a comment that they're aliases so nobody "dedupes" the wrong one.

voyageai/local/model_registry.py
fzowl: Fixed in 006b8d7. \`__new__\` now builds a local instance, sets \`_cache\`/\`_cache_lock\` on it, and assigns \`cls._instance\` only as the final step — so the lock-free outer \`is None\` check can never observe a published-but-half-built singleton.

voyageai/_base.py
fzowl: Fixed in 006b8d7. Construction now catches \`(AuthenticationError, OSError)\` (FileNotFoundError is an OSError subclass), so a set-but-stale \`VOYAGE_API_KEY_PATH\` falls back to keyless. Verified: with \`VOYAGE_API_KEY\` unset and \`VOYAGE_API_KEY_PATH=/nonexistent/key.txt\`, \`Client()\` now constructs with \`api_key=None\` instead of raising.

voyageai/local/sentence_transformer_backend.py
fzowl: Fixed in 006b8d7. \`normalize_embeddings=True\` is now set unconditionally rather than only on the truncated branch, so full-dim output is unit-norm by guarantee instead of relying on the model self-normalizing. \`test_truncated_dimensions_are_unit_norm\` (incl. dim 2048) passes.

voyageai/local/sentence_transformer_backend.py
fzowl: Good catch. The per-call \`processing_kwargs\` override that \`encode()\` relies on for \`truncation=False\` was added in sentence-transformers 5.5.0 (PR #3753), but the floor was \`>=5.0.0\`. Bumped to \`>=5.5.0\` in 8dad5ee so the kwarg can no longer be silently dropped on older 5.x.

voyageai/__init__.py
fzowl: Simplified in 006b8d7. Collapsed the \`_LazyLocalFlag\`/\`_Module\`/\`_module_proxy\` chain into a plain module \`__getattr__\` that imports and calls \`_is_local_available()\` on access to \`HAS_LOCAL\`.

tests/test_client_local_async.py
fzowl: Fixed in 006b8d7. Extracted the dep-probe into \`tests/_local_test_utils.real_deps_available\`; both local test modules import it. Kept it \`find_spec\`-based so it stays an independent signal for \`test_has_local_constant\`.

voyageai/local/helpers.py
fzowl: Fixed in 006b8d7. The \`get_local_backend\` docstring now states that only the underlying model is cached (process-wide, keyed by model:device via ModelCache) and a fresh backend wrapper is built per call.

voyageai/local/sentence_transformer_backend.py
fzowl: Fixed in 006b8d7. Collapsed the passthrough — \`_count_tokens_with_prefix\` is gone and \`count_tokens\` is the single public method (encode() calls it). Kept the per-text prefixed count intentionally: it mirrors the instruction prefix the model actually tokenizes, so the reported \`total_tokens\` matches server-side usage accounting.

voyageai/local/model_registry.py
fzowl: Fixed in 006b8d7. \`loader\` is now typed \`Callable[[], Any]\` (\`Callable\` added to the \`typing\` import).

voyageai/_base.py
fzowl: Keeping the explicit \`_require_api_key()\` gate rather than a raising \`api_key\` property. All four API-backed methods call it at their boundary, and \`count_usage\`/\`tokenize\` are intentionally keyless. A property that raises on *access* would make those keyless paths fragile — any incidental \`self.api_key\` read would throw — and it hides the check from the call site. The explicit one-liner is easy to see in review when a method is added. Noted as something to watch as the surface grows.

voyageai/local/model_registry.py
fzowl: Keeping both checks intentionally — they encode different semantics. \`supported_precisions\` is the validation allowlist and raises \`InvalidRequestError\` (the user passed an invalid \`output_dtype\`). \`_UNSUPPORTED_QUANTIZED_DTYPES\` raises \`NotImplementedError\` for dtypes that are valid against the hosted API but not yet supported locally — a capability gap, not a user error. int8/uint8 are checked first, so they always take the NotImplementedError path. The two cannot "silently disagree" into accepting a bad dtype: if they ever drift, the worst case is a less-precise error *message* (InvalidRequestError vs NotImplementedError), never silent acceptance. The \`float32\`/\`float\` pair are documented aliases (both map to float32, matching the API treating \`float\` as the default).

fzowl:
Thanks for the thorough pass, @jasonqiu95. Addressed the latest round (commits 006b8d7 + 8dad5ee) and rebased onto `main` (resolved the import conflicts in `client.py`/`client_async.py`):

**Fixed**
- `ModelCache.__new__` publish-before-init race — instance is fully initialized before `cls._instance` is published.
- Keyless construction now catches `OSError` (stale `VOYAGE_API_KEY_PATH`).
- Embeddings always normalized (full-dim unit-norm by guarantee, not by implicit model behavior).
- `sentence-transformers` floor bumped to `>=5.5.0` (first release with per-call `processing_kwargs`, which `truncation=False` relies on).
- Simplified `HAS_LOCAL` to a plain module `__getattr__`.
- `loader: Callable[[], Any]`, corrected `get_local_backend` docstring, collapsed the `count_tokens` passthrough, de-duplicated the test dep-probe into `tests/_local_test_utils`.

**Kept (with rationale in the inline threads)**
- The explicit `_require_api_key()` gate over a raising `api_key` property.
- The dual int8/uint8 gate — the two checks carry different error semantics (`InvalidRequestError` vs `NotImplementedError`) and cannot silently disagree into accepting a bad dtype.

Local unit + integration tests pass locally.

.github/workflows/ci.yaml
```diff
@@ -7,17 +7,88 @@ on:
...
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
```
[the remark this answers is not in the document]
fzowl: Fixed in 8683b86 — pinned this action (and the others across all three jobs) to a full commit SHA with a version comment, removing the mutable-tag risk.

.github/workflows/ci.yaml
```diff
@@ -7,17 +7,88 @@ on:
...
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
     env:
       VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
-      POETRY_HOME: "/opt/poetry"
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
```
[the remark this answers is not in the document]
fzowl: Fixed in 8683b86 — pinned this action (and the others across all three jobs) to a full commit SHA with a version comment, removing the mutable-tag risk.

.github/workflows/ci.yaml
```diff
@@ -7,17 +7,88 @@ on:
...
+  local-model-tests:
+    name: Local Model Tests
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
```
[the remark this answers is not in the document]
fzowl: Fixed in 8683b86 — pinned this action (and the others across all three jobs) to a full commit SHA with a version comment, removing the mutable-tag risk.

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,197 @@
...
+        # only on the truncated branch) makes full-dim norm an explicit
+        # guarantee instead of relying on the model self-normalizing.
+        encode_kwargs = {"normalize_embeddings": True}
+        if dimension != self.config.default_dimension:
+            encode_kwargs["truncate_dim"] = dimension
+        if precision:
+            encode_kwargs["precision"] = precision
+
+        # Wire truncation through to the model via processing_kwargs
+        if not truncation:
```
jasonqiu95: **[Correctness/Parity] `truncation=False` on over-length input doesn't raise `InvalidRequestError` the way the hosted API does.**

The API contract (see `tests/test_client.py::test_client_embed_context_length`) is: `truncation=True` truncates and succeeds, `truncation=False` on input exceeding the context limit raises `InvalidRequestError`. Here, `truncation=False` only sets `processing_kwargs={"text": {"truncation": False}}` and feeds the un-truncated sequence to the model — so an over-length input either crashes deep in the model (RuntimeError/IndexError on the position embedding) or silently degrades, never the `InvalidRequestError` callers catch on the API path.

Suggested fix — validate length up front when truncation is disabled:
```python
if not truncation:
    prompt = self._get_prompt_for_input_type(input_type)
    for text in texts:
        full = f"{prompt}{text}" if prompt else text
        if len(self._tokenizer.encode(full, add_special_tokens=True)) > self.config.max_tokens:
            raise InvalidRequestError(
                f"Input exceeds the {self.config.max_tokens}-token context length "
                "and truncation is disabled."
            )
    encode_kwargs["processing_kwargs"] = {"text": {"truncation": False}}
```

voyageai/local/helpers.py
```diff
@@ -0,0 +1,77 @@
...
+        output_dimension: Output embedding dimension.
+
+    Returns:
+        EmbeddingsObject with embeddings and token count.
+    """
+    # The API path accepts a bare string and returns a list with one embedding
+    # (embeddings == [[...]]). Normalize here so the local path matches; without
+    # this, encode() would return a 1-D array and embeddings[0] would be a float.
+    if isinstance(texts, str):
+        texts = [texts]
```
jasonqiu95, line 58: **[Correctness/Parity] Empty / non-list input diverges from the API.**

The bare-string normalization here is good, but two adjacent cases still diverge from the hosted API:
- `client.embed([], model="voyage-4-nano")` → `encode([])` returns an empty array, so this returns `embeddings=[]`, `total_tokens=0` with no error, whereas the API rejects empty input with `InvalidRequestError`.
- `client.embed(None, ...)` → `isinstance(None, str)` is False, so `None` flows into `encode()` and `count_tokens()` does `for text in texts` → `TypeError: 'NoneType' is not iterable`, instead of a clean `InvalidRequestError`.

Suggested fix — validate right after the str-wrap:
```python
if isinstance(texts, str):
    texts = [texts]
if not texts:
    raise InvalidRequestError("inputs must not be empty.")
```
(also rejects `None`, since `not None` is True).

voyageai/local/helpers.py
```diff
@@ -0,0 +1,77 @@
...
+
+    backend = get_local_backend(model)
+
+    embeddings_array, total_tokens = backend.encode(
```
jasonqiu95, line 78: **[Parity] No upper bound on batch size locally — the API rejects oversized batches.**

`tests/test_client.py::test_client_embed_batch_size` shows the API raises `InvalidRequestError` once a batch exceeds the documented limit. The local path accepts an arbitrarily large `texts` list and encodes all of it in one shot, so instead of a clean error a too-large batch just degrades to a slow run or an OOM. Same `client.embed(...)` call, different failure mode between hosted and local.

Consider enforcing the model's documented batch limit (store it on `LocalModelConfig`) and raising `InvalidRequestError` when exceeded, so callers get the same error on both paths.

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,197 @@
...
+            input_type: "query", "document", or None.
+
+        Returns:
+            Total token count.
+        """
+        prompt = self._get_prompt_for_input_type(input_type)
+        total = 0
+        for text in texts:
+            full_text = f"{prompt}{text}" if prompt else text
+            encoded = self._tokenizer.encode(full_text, add_special_tokens=True)
```
jasonqiu95: **[Correctness] `total_tokens` overcounts when `truncation=True` truncates an over-length input.**

`model.max_seq_length` is set to `max_tokens` (32768), so with the default `truncation=True` the model embeds only the first 32768 tokens of an over-length input. But `count_tokens()` (line 195) re-tokenizes the *full* text with no length cap, so `result.total_tokens` reports more tokens than were actually embedded — diverging from the API, which reports the processed (post-truncation) count. (Distinct from the already-fixed prefix-counting issue.)

Suggested fix — cap the per-text count when truncation is on:
```python
n = len(self._tokenizer.encode(full_text, add_special_tokens=True))
total += min(n, self.config.max_tokens) if truncation else n
```
(requires threading `truncation` into `count_tokens`).

voyageai/_base.py
```diff
@@ -60,16 +60,43 @@ def __init__(
         timeout: Optional[float] = None,
         base_url: Optional[str] = None,
     ) -> None:
-        self.api_key = api_key or default_api_key()
+        # API key is optional - allow None for local-only usage
+        try:
+            self.api_key = api_key or default_api_key()
+        except (AuthenticationError, OSError):
```
jasonqiu95: **[Correctness] `except (AuthenticationError, OSError)` misses `UnicodeDecodeError` from a non-UTF-8 key file.**

`default_api_key()` opens `VOYAGE_API_KEY_PATH` in text mode (`open(path, "rt")`) and `.read()`s it. A file with non-UTF-8/binary content raises `UnicodeDecodeError`, which is a subclass of `ValueError` — not `OSError`. So a stale/garbage key file crashes construction with `UnicodeDecodeError`, contradicting this block's stated intent of falling back to keyless.

Suggested fix:
```python
except (AuthenticationError, OSError, ValueError):
```

voyageai/local/model_registry.py
```diff
@@ -0,0 +1,170 @@
...
+        Thread-safe: holds the lock for the full check-load-store sequence.
+
+        Args:
+            key: Cache key.
+            loader: Callable to load the model if not cached.
+
+        Returns:
+            The cached or newly loaded model.
+        """
+        with self._cache_lock:
```
jasonqiu95: **[Correctness/Concurrency] `get_or_load` holds the cache lock across the whole multi-second model load, serializing unrelated cache access.**

The full check-load-store runs under `_cache_lock`, and `get`/`set`/`clear` take the same lock. So while one thread cold-loads `voyage-4-nano`, any other thread asking for an *already-cached* (different) key blocks for the entire load. The async path makes this concrete: concurrent `AsyncClient.embed(...)` calls dispatch `embed_local` to several `asyncio.to_thread` workers, which all stall behind the first cold load instead of proceeding with already-loaded models.

Suggested fix — use a per-key lock so only same-key loads serialize:
```python
with self._cache_lock:
    model = self._cache.get(key)
    if model is not None:
        return model
    key_lock = self._key_locks.setdefault(key, threading.Lock())
with key_lock:                       # load outside the global lock
    model = self._cache.get(key)
    if model is None:
        model = loader()
        self.set(key, model)
    return model
```

voyageai/local/sentence_transformer_backend.py
```diff
@@ -0,0 +1,197 @@
...
+
+        # Wire truncation through to the model via processing_kwargs
+        if not truncation:
+            encode_kwargs["processing_kwargs"] = {
+                "text": {"truncation": False},
+            }
+
+        # Route based on input_type using prompt_name
+        if input_type == "query":
+            embeddings = self.model.encode_query(texts, **encode_kwargs)
```
jasonqiu95: **[Concurrency — please verify] Concurrent async embeds call `encode()` on one shared cached model from multiple threads.**

The PR adds `test_concurrent_local_embeddings`, which `asyncio.gather`s several `embed` calls; each routes through `asyncio.to_thread(embed_local, ...)`, and every worker thread calls `self.model.encode*()` on the *same* `ModelCache`-singleton `SentenceTransformer`. PyTorch inference forward passes are usually safe, but `SentenceTransformer.encode` is not contractually thread-safe (it can sort/batch and touch shared module state), so concurrent calls could in principle interleave — and the new test would still pass even if they did.

Defensive options: a per-model inference lock around the `encode*` calls, or documenting that the cached model must be accessed serially. Worth confirming against the sentence-transformers version pinned here before relying on it.

jasonqiu95 (changes requested):

voyageai/local/sentence_transformer_backend.py
fzowl: Fixed in 0001920. The \`encode*\` forward passes now run under a per-model inference lock (\`ModelCache.get_inference_lock\`, keyed by \`model:device\`), so concurrent async embeds that share the cached model serialize their forward passes instead of racing.

voyageai/local/model_registry.py
fzowl: Fixed in 0001920. \`get_or_load\` now takes a **per-key** load lock: a lock-free fast-path read, then (on miss) a per-key lock with a re-check before loading. A cold load of one key no longer blocks callers of a different, already-cached key.

voyageai/_base.py
fzowl: Fixed in 0001920. Construction now also catches \`UnicodeDecodeError\`, so a non-UTF-8 \`VOYAGE_API_KEY_PATH\` falls back to keyless. Added a test that points the path at a binary file and asserts \`api_key is None\`.

voyageai/local/sentence_transformer_backend.py
fzowl: Fixed in 0001920. \`count_tokens\` now takes \`truncation\` and caps each text at \`max_tokens\` when truncation is on, so \`total_tokens\` reports the processed count like the API. Added a test asserting \`total_tokens <= 32768\` for an over-length input.

voyageai/local/helpers.py
fzowl: Fixed in 0001920. \`embed_local\` now raises \`InvalidRequestError\` on empty input (and \`None\`, since \`not None\` is True) right after the str-wrap, before any model load. Added tests for \`[]\` and \`None\`.

voyageai/local/sentence_transformer_backend.py
fzowl: Fixed in 0001920. With \`truncation=False\`, over-length input is now rejected up front with \`InvalidRequestError\` (validated against \`max_tokens\`) instead of feeding an over-length sequence to the model. Added an integration test.

voyageai/local/helpers.py
fzowl: Thanks for this — I looked at it closely and decided to leave it as-is for now, hope that's alright.

The API's batch-size limit is a server-side constraint; on the local path the practical bound is host memory, so a hard cap mirroring the hosted number would reject legitimate local usage (e.g. large offline batches) without a real correctness gain — and there's no first-class "documented local batch limit" to key it off yet. I'd rather document the difference than raise an artificial error, and I'm happy to revisit if we publish an explicit local limit or start seeing real OOM reports.

Much appreciated the parity pass overall — the sibling points (empty/None input, truncation=False, token counting under truncation) were spot-on and are fixed in 0001920.

fzowl: Update: on reflection I reversed my earlier decline and implemented this in 0c24b4f. You're right that it's a documented contract, not just a server-side detail — the embeddings docs cap the input list at 1000, and tests/test_client.py::test_client_embed_batch_size rejects an oversized batch with InvalidRequestError. That's the same input-validation parity as the empty/None and truncation cases, so enforcing it locally is the consistent choice.

Added max_batch_size (=1000) to LocalModelConfig and embed_local now raises InvalidRequestError before loading the model when the list exceeds it, with a test. Thanks for pushing back on this one.

.github/workflows/ci.yaml
```diff
@@ -20,9 +24,81 @@ jobs:
       max-parallel: 1
```
jasonqiu95, line 24: **[CI] `max-parallel: 1` here is justified by a comment copied from `integration-tests`, but it's false for this job — it needlessly serializes the matrix ~6×.**

Root cause: the comment claims this job *"runs the full suite against the live API (incl. token-heavy voyage-multimodal-3.5)"* and must serialize to stay under the TPM rate limit. But `unit-tests-no-local-deps` sets **no** `VOYAGE_API_KEY`, installs only `--extras video` (no local deps), and runs just `tests/test_client_local.py tests/test_client_local_async.py`. With no deps and no key every `@pytest.mark.integration` test skips, so this job makes **zero** API calls — the TPM rationale doesn't apply. The block was cloned verbatim from `integration-tests` (lines 93–96), where it *is* correct.

Effect: the 6-version matrix runs one Python at a time for no reason (~6× slower than it needs to be), and the inaccurate comment will mislead future maintainers into thinking this job is API-bound.

Suggested fix — drop `max-parallel: 1` and the copied comment from this job so the offline matrix runs in parallel, and keep the serialization only on `integration-tests` (the job that actually hits the live API):

```yaml
  unit-tests-no-local-deps:
    name: Unit Tests (no local deps)
    runs-on: ubuntu-22.04
    strategy:
      fail-fast: false
      matrix:
        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
```

jasonqiu95 (approved):

.github/workflows/ci.yaml
fzowl: Thanks @jasonqiu95, good catch — you're right that this job sets no VOYAGE_API_KEY and installs no local deps, so it makes zero API calls and the TPM/max-parallel justification (copied from integration-tests) doesn't apply here. Since it's a CI-perf/comment cleanup and not a functional issue, I've tracked it separately in #68 so we can land this PR without a re-review cycle. Resolving this thread.

PR #60, opened 2026-06-01
Migrate to Pydantic v2 APIs for Python 3.14 support

fzowl:
The previous fix (#46) swapped the pydantic import order but still fell back to the broken pydantic.v1 path since Extra was removed in v2.

- Remove pydantic.v1 fallback and Extra import
- Replace Extra.forbid with string "forbid"
- Replace min_items with min_length (v2 rename)
- Replace .parse_obj() with .model_validate()
- Replace .dict() with .model_dump()
- Bump pydantic lower bound to >=2.7.4 (already required transitively by langchain-text-splitters via langchain-core)
- Bump version to 0.4.0 (breaking change for downstream consumers pinning pydantic v1)

Fixes #36

Reopened from #50 — moved to upstream branch for CI secret access.

---

**Previous discussion from #50:**

> **@jasonqiu95** (on `pyproject.toml`):
> Going from pydantic >=1.10.8 to pydantic >=2.7.4 is a breaking change for any downstream consumer pinning v1. So it should be 0.4.0. We should also note this clearly that we switched to pydantic v2 in the release note.

> **@fzowl**: Good call — bumped to 0.4.0.

> **@jasonqiu95** (on `tests/test_client_async.py`):
> ~~this change seems unrelated could you move it to another PR and explain why it's needed?~~
> nvm I figured out this is to make your fork's CI green right? Also I trust that the main repo's CI run will run these tests with an actual key right?

> **@fzowl**: Yeah, exactly!

> **@abgros**: @fzowl what's the status of this PR?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "voyageai"
-version = "0.3.8"
+version = "0.3.9"
 description = ""
 authors = ["Yujie Qian <yujieq@voyageai.com>"]
 readme = "README.md"
@@ -16,7 +16,7 @@ numpy = [
 ]
 aiolimiter = "*"
 pillow = "*"
-pydantic = ">=1.10.8"
+pydantic = ">=2.7.4"
 tokenizers = ">=0.14.0"
 langchain-text-splitters = ">=0.3.8"
 ffmpeg-python = "*"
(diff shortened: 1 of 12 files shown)


jasonqiu95 (approved):
lgtm

PR #61, opened 2026-06-01
Fix/circular import error 42

fzowl:
Resolving https://github.com/voyage-ai/voyageai-python/issues/42

Reopened from #51 — moved to upstream branch for CI secret access.

### DIFF
--- a/tests/test_import_order.py
+++ b/tests/test_import_order.py
@@ -1,0 +1,40 @@
+"""Regression tests for circular import (issue #42).
+
+Importing submodules directly (without a prior ``import voyageai``) used to
+trigger an ``ImportError`` / ``AttributeError`` because some modules accessed
+``voyageai.error.X`` via late-binding attribute access on the top-level package,
+which hadn't finished initializing yet.
+
+Each test below spawns a **fresh** Python subprocess that imports a single
+submodule — this is the only reliable way to reproduce the original failure,
+because in-process importlib tricks are defeated by the module cache.
+"""
+
+import subprocess
+import sys
+
+import pytest
+
+# Each entry is (human-readable label, Python import statement).
+_IMPORT_CASES = [
+    ("voyageai.util", "from voyageai.util import default_api_key"),
+    ("voyageai.embeddings_utils", "from voyageai.embeddings_utils import get_embeddings"),
+    (
+        "voyageai.api_resources.api_requestor",
+        "from voyageai.api_resources.api_requestor import APIRequestor",
+    ),
+]
+
+
+@pytest.mark.parametrize("label,import_stmt", _IMPORT_CASES, ids=[c[0] for c in _IMPORT_CASES])
+def test_no_circular_import_error(label: str, import_stmt: str) -> None:
+    """Importing *label* in a fresh interpreter must not raise ImportError."""
+    result = subprocess.run(
+        [sys.executable, "-c", import_stmt],
+        capture_output=True,
+        text=True,
+        timeout=30,
+    )
+    assert result.returncode == 0, (
+        f"Importing {label} failed (exit {result.returncode}).\n" f"stderr:\n{result.stderr}"
+    )
--- a/tests/test_import_order.py
+++ b/tests/test_import_order.py
@@ -10,10 +10,19 @@ submodule — this is the only reliable way to reproduce the original failure,
 because in-process importlib tricks are defeated by the module cache.
 """
 
+import pathlib
 import subprocess
 import sys
 
 import pytest
+
+import voyageai
+
+# Modules that previously evaluated ``voyageai.error.X`` via late-binding
+# attribute access at import time (the root cause of issue #42). They must now
+# import error classes directly (``from voyageai.error import X``).
+_PATCHED_MODULES = ["util.py", "embeddings_utils.py", "_base.py"]
+_PKG_DIR = pathlib.Path(voyageai.__file__).parent
 
 # Each entry is (human-readable label, Python import statement).
 _IMPORT_CASES = [
@@ -37,4 +46,21 @@ def test_no_circular_import_error(label: str, import_stmt: str) -> None:
     )
     assert result.returncode == 0, (
         f"Importing {label} failed (exit {result.returncode}).\n" f"stderr:\n{result.stderr}"
+    )
+
+
+@pytest.mark.parametrize("filename", _PATCHED_MODULES)
+def test_no_late_bound_error_access(filename: str) -> None:
+    """The subprocess test above can pass even against the pre-fix code, because
+    importing any submodule first runs ``voyageai/__init__.py`` to completion,
+    which binds ``voyageai.error`` before the late access is reached. This static
+    check actually fails on the buggy code and passes on the fix: it asserts the
+    previously-affected modules contain no module-scope ``voyageai.error.X``
+    attribute access.
+    """
+    source = (_PKG_DIR / filename).read_text()
+    assert "voyageai.error." not in source, (
+        f"{filename} uses late-bound 'voyageai.error.X' access (issue #42). "
+        "Import the error class directly instead, e.g. "
+        "'from voyageai.error import RateLimitError'."
     )
--- a/tests/test_import_order.py
+++ b/tests/test_import_order.py
@@ -15,7 +15,6 @@ import subprocess
 import sys
 
 import pytest
-
 import voyageai
 
 # Modules that previously evaluated ``voyageai.error.X`` via late-binding
--- a/voyageai/__init__.py
+++ b/voyageai/__init__.py
@@ -25,6 +25,12 @@ from voyageai.api_resources import (
     MultimodalEmbedding,
     Reranking,
 )
+
+# Bind the ``error`` submodule onto the package eagerly and explicitly. Without
+# this, ``voyageai.error`` is only bound as an incidental side effect of another
+# submodule's import, which made late-bound ``voyageai.error.X`` accesses during
+# module load order-dependent and could raise AttributeError (issue #42).
+from voyageai import error
 from voyageai.chunking import default_chunk_fn
 from voyageai.client import Client
 from voyageai.client_async import AsyncClient
--- a/voyageai/_base.py
+++ b/voyageai/_base.py
@@ -10,7 +10,6 @@ import PIL.Image
 from huggingface_hub import hf_hub_download
 
 import voyageai
-from voyageai.error import InvalidRequestError
 from voyageai.object import EmbeddingsObject, RerankingObject
 from voyageai.object.contextualized_embeddings import ContextualizedEmbeddingsObject
 from voyageai.object.multimodal_embeddings import (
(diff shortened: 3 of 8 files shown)


jasonqiu95:
**Code review — looks good. ✅**

Reviewed the fix for #42 (converting late-binding `voyageai.error.X` attribute access → explicit `from voyageai.error import X`). Verified empirically, not just by inspection:

- **Behaviorally identical.** Confirmed the top-level-imported names are the *same class objects* as `voyageai.error.X` (`is` check) in all three modules, so both `raise X(...)` and `retry_if_exception_type(X)` are unchanged.
- **No new cycle.** `error.py` only does `import voyageai` with no back-dependency on `_base` / `util` / `embeddings_utils`, so `from voyageai.error import X` can't form a new cycle. Tested several adversarial entry points (`import voyageai`, `from voyageai.util import default_api_key` first, `import voyageai.embeddings_utils` first, etc.) in fresh interpreters — all succeed.
- **Complete.** Zero remaining `voyageai.error.X` attribute-access call sites after the change. The surviving `import voyageai.error as error` / `from voyageai import error` forms are direct submodule imports and are *not* susceptible to the #42 lazy-import bug, so they're correctly left alone.
- **No dead imports.** `import voyageai` is still needed in all three files (`VOYAGE_EMBED_*`, `api_key`, `Embedding`, `log`).

**Optional follow-up (non-blocking):** there's no regression test guarding against reintroduction. The existing tests run after a full `import voyageai`, so they wouldn't catch a #42 regression. A lightweight guard — import each affected module in a fresh subprocess via `importlib` and assert no `ImportError` — would lock this in.

jasonqiu95 (commented):
## Code review — circular import fix (#42)

The **source fix is correct and I verified it**: on this branch no `voyageai.error.X` late-binding accesses remain (`grep` is clean), all three changed modules plus `_base` import cleanly in fresh interpreters, and `from voyageai.error import RateLimitError` resolves to the *same* class object as the old `voyageai.error.RateLimitError`, so `tenacity`'s `retry_if_exception_type` and user `except` clauses keep matching. 👍

My concerns are all about the **regression test**, which doesn't actually guard the bug it targets, plus an **altitude** question about how deep the fix goes. Details inline; the most important one first:

**1. The regression test passes against the *pre-fix* (buggy) code — so it guards nothing.** I checked out the pre-PR versions of `util.py` / `embeddings_utils.py` / `_base.py` (the ones still using `voyageai.error.X`), kept this new test file, and ran it: **all 3 cases pass.** The reason: any `from voyageai.X import ...` first runs `voyageai/__init__.py` to completion, and there `from voyageai.api_resources import (...)` (line 22) executes `api_requestor`'s `from voyageai import error` — binding `voyageai.error` onto the package — *before* `embeddings_utils` is imported (line 31). So the old decorator-arg `voyageai.error.RateLimitError` was already resolvable, and the test can't distinguish fixed from broken code. To actually reproduce #42 you'd need to force the unbound-ordering condition (e.g. import a submodule with `voyageai.error` not yet bound) or assert statically that the modules contain no module-scope `voyageai.error.X` references.

**2. Altitude — the structural fragility is left in place.** This PR took issue #42's per-site "Option 2" but not "Option 1" (`from voyageai import error` in `__init__.py`). `voyageai/__init__.py` *still* never explicitly imports the `error` submodule; `voyageai.error` is bound only as an incidental side effect of `api_requestor`'s import during line 22. The three patched files no longer depend on that ordering, but any *new* module-load-time `voyageai.error.X` (or a reordering of `__init__.py`) reintroduces the same intermittent failure. A one-line `from voyageai import error` near the top of `__init__.py` would make the whole class of bug impossible.

**3. Incomplete migration (consistency).** `voyageai/client.py` and `voyageai/client_async.py` still use `import voyageai.error as error` + `error.RateLimitError` etc. These are safe (the alias binds eagerly and the refs are inside `_make_retry_controller`, evaluated at call time), but they're the same indirect-access pattern this PR set out to eliminate, so the codebase is left half-migrated.

**4. Heads-up (pre-existing, out of scope):** `error.py:63` references `voyageai.api_resources.error_object.ErrorObject`, but no `error_object` module / `ErrorObject` class exists in the repo. Constructing any `VoyageError` with a structured `json_body={'error': {...}}` raises `AttributeError`, masking the real API error. Not introduced here, but directly adjacent to this PR's theme — worth a follow-up.

Nothing here blocks the source change; the test is the part I'd want strengthened before merge.

tests/test_import_order.py
```diff
@@ -0,0 +1,40 @@
...
+@pytest.mark.parametrize("label,import_stmt", _IMPORT_CASES, ids=[c[0] for c in _IMPORT_CASES])
+def test_no_circular_import_error(label: str, import_stmt: str) -> None:
+    """Importing *label* in a fresh interpreter must not raise ImportError."""
+    result = subprocess.run(
+        [sys.executable, "-c", import_stmt],
+        capture_output=True,
+        text=True,
+        timeout=30,
+    )
+    assert result.returncode == 0, (
```
jasonqiu95, line 46: **This test passes against the un-fixed code, so it doesn't guard #42.** Reverting the three source files to their old `voyageai.error.X` versions and running this test gives `3 passed` — importing any submodule first runs `voyageai/__init__.py`, where `api_resources` (line 22) binds `voyageai.error` before `embeddings_utils` loads, so the late-bound access was always resolvable on these paths.

A static check actually fails on the buggy code and passes on the fix (verified both ways):

```python
import pathlib
import pytest
import voyageai

# Modules that used to evaluate ``voyageai.error.X`` at import time (#42).
_PATCHED = ["util.py", "embeddings_utils.py", "_base.py"]
_PKG = pathlib.Path(voyageai.__file__).parent


@pytest.mark.parametrize("filename", _PATCHED)
def test_no_late_bound_error_access(filename):
    source = (_PKG / filename).read_text()
    assert "voyageai.error." not in source, (
        f"{filename} uses voyageai.error.X (late binding); import the error "
        "class directly (from voyageai.error import X) to avoid #42."
    )
```

jasonqiu95 (changes requested):

tests/test_import_order.py
```diff
@@ -0,0 +1,65 @@
+"""Regression tests for circular import (issue #42).
+
+Importing submodules directly (without a prior ``import voyageai``) used to
+trigger an ``ImportError`` / ``AttributeError`` because some modules accessed
+``voyageai.error.X`` via late-binding attribute access on the top-level package,
+which hadn't finished initializing yet.
+
+Each test below spawns a **fresh** Python subprocess that imports a single
+submodule — this is the only reliable way to reproduce the original failure,
+because in-process importlib tricks are defeated by the module cache.
```
jasonqiu95, line 10: **The subprocess test does not catch the bug it's named for, and this docstring is misleading.**

Root cause: this module docstring claims spawning a fresh subprocess "is the only reliable way to reproduce the original failure." That's inaccurate, and it directly contradicts the docstring of `test_no_late_bound_error_access` below (lines 53–58), which correctly states *"The subprocess test above can pass even against the pre-fix code."*

I verified this empirically — running this exact test file against the pre-fix `main` source:

```
test_no_circular_import_error[voyageai.util]                PASSED
test_no_circular_import_error[voyageai.embeddings_utils]    PASSED
test_no_circular_import_error[voyageai.api_resources...]    PASSED
test_no_late_bound_error_access[util.py]                    FAILED
test_no_late_bound_error_access[embeddings_utils.py]        FAILED
test_no_late_bound_error_access[_base.py]                   FAILED
```

The subprocess cases pass on buggy code; only the static check is a real regression guard. The danger is a future maintainer trusting the subprocess test (per this docstring) and deleting the static one — losing all protection against #42.

Suggested fix — describe the two checks accurately:

```suggestion
Two complementary checks guard against a regression of issue #42:

* ``test_no_late_bound_error_access`` is the real guard: a static check that
  fails on the buggy code and passes on the fix.
* ``test_no_circular_import_error`` spawns a fresh subprocess per submodule to
  ensure a direct import doesn't raise. Note it passes even against the pre-fix
  code (see its docstring), so it's a weaker, complementary smoke test.
```

tests/test_import_order.py
```diff
@@ -0,0 +1,65 @@
...
+import sys
+
+import pytest
+import voyageai
+
+# Modules that previously evaluated ``voyageai.error.X`` via late-binding
+# attribute access at import time (the root cause of issue #42). They must now
+# import error classes directly (``from voyageai.error import X``).
+_PATCHED_MODULES = ["util.py", "embeddings_utils.py", "_base.py"]
+_PKG_DIR = pathlib.Path(voyageai.__file__).parent
```
jasonqiu95, line 24: **The static check reads the *installed* package, not the working-tree source — risking a false pass.**

Root cause: `_PKG_DIR` is derived from `voyageai.__file__`, i.e. wherever the `voyageai` package was imported from. The check on line 60 then does `(_PKG_DIR / filename).read_text()`. Under an editable install this is the working tree, but under a non-editable install (e.g. CI does `pip install .` / `poetry install` into site-packages, then runs the tests) `_PKG_DIR` points at the installed copy. The guard would then validate a stale snapshot rather than the code under review — a false pass (or false fail) decoupled from the PR's actual changes.

Suggested fix — anchor to the repo source relative to the test file so the check always reads what's under review:

```suggestion
_PKG_DIR = pathlib.Path(__file__).resolve().parent.parent / "voyageai"
```

tests/test_import_order.py
```diff
@@ -0,0 +1,65 @@
...
+import subprocess
+import sys
+
+import pytest
+import voyageai
+
+# Modules that previously evaluated ``voyageai.error.X`` via late-binding
+# attribute access at import time (the root cause of issue #42). They must now
+# import error classes directly (``from voyageai.error import X``).
+_PATCHED_MODULES = ["util.py", "embeddings_utils.py", "_base.py"]
```
jasonqiu95, line 23: **Minor / optional: the regression guard skips two files this PR also converted.**

Root cause: `client.py` and `client_async.py` were also migrated off `voyageai.error.X` in this PR (they used the safe `import voyageai.error as error` alias, so they weren't part of the original #42 crash), but they're absent from `_PATCHED_MODULES`. Since the codebase now follows a "no late-bound `voyageai.error.X`" convention everywhere, a future reintroduction of the pattern in those two files would not be caught by this static guard.

Suggested fix — add them so the guard covers every migrated module:

```suggestion
_PATCHED_MODULES = ["util.py", "embeddings_utils.py", "_base.py", "client.py", "client_async.py"]
```

(Note: a whole-package glob would be more future-proof, but the substring check on line 61 also matches comments/strings — e.g. the explanatory comment in `__init__.py` contains `voyageai.error.` — so a curated list is the pragmatic choice here.)

jasonqiu95 (commented):
Reviewed the circular-import fix (recall pass). The core change is correct — verified in a worktree:

- `error_object.ErrorObject` never existed in repo history, so on `main`, constructing any `VoyageError` whose `json_body` contains an `"error"` dict raises `AttributeError`. Switching to `VoyageResponse.construct_from` is a genuine fix, and it produces the expected dict-shaped `.error`.
- All 6 import-order tests pass; the direct-import refactor across the 5 modules is sound.

No correctness bugs found. The findings below are about the new test file's accuracy and robustness. Inline comments give root cause + suggested fix.

jasonqiu95 (approved):

PR #64, opened 2026-06-20
Add test coverage reporting to CI

fzowl:
## Summary
- Add `pytest-cov` to test dependencies
- Run tests with `--cov=voyageai --cov-report=term-missing` for visibility into coverage gaps
- Upload `coverage.xml` as artifact on Python 3.12 runs for downstream tooling

## Local results (Python 3.13, 209 tests passing)

| Module | Coverage |
|--------|----------|
| Overall | **79%** |
| `__init__.py` | 100% |
| `client.py` / `client_async.py` | 90% |
| `api_requestor.py` | 69% |
| `video_utils.py` | 72% |
| `embeddings_utils.py` | 36% |
| `response.py` | 52% |

## Test plan
- [ ] CI runs with coverage report visible in logs
- [ ] Coverage XML artifact uploaded for 3.12
- [ ] No test regressions

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,9 +20,9 @@ jobs:
       POETRY_HOME: "/opt/poetry"
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install poetry
@@ -44,14 +44,17 @@ jobs:
       run: $POETRY_HOME/bin/poetry run pytest --cov=voyageai --cov-report=term-missing --cov-report=xml:coverage.xml
     - name: Enforce coverage threshold
       # Gate once, on a single canonical version, instead of redundantly on
-      # every matrix entry. Threshold tracks the coverage this PR achieves.
+      # every matrix entry. The full suite (including the live-API integration
+      # tests, which run because VOYAGE_API_KEY is set above) measures ~93% here;
+      # the gate is set to 90% to leave headroom so a single flaky API test or a
+      # small uncovered addition doesn't red the build.
       if: matrix.python-version == '3.12'
-      run: $POETRY_HOME/bin/poetry run coverage report --fail-under=93
+      run: $POETRY_HOME/bin/poetry run coverage report --fail-under=90
     - name: Upload coverage report
       # always() so the report is still uploaded when the threshold gate fails,
       # which is exactly when it's most useful for diagnosis.
       if: always() && matrix.python-version == '3.12'
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4.4.3
       with:
         name: coverage-report
         path: coverage.xml
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,5 +40,11 @@ jobs:
         ffmpeg -version
     - name: Install package (including video extra for multimodal tests)
       run: $POETRY_HOME/bin/poetry install --extras video
-    - name: Run tests
-      run: $POETRY_HOME/bin/poetry run pytest
+    - name: Run tests with coverage
+      run: $POETRY_HOME/bin/poetry run pytest --cov=voyageai --cov-report=term-missing --cov-report=xml:coverage.xml
+    - name: Upload coverage report
+      if: matrix.python-version == '3.12'
+      uses: actions/upload-artifact@v4
+      with:
+        name: coverage-report
+        path: coverage.xml
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,9 +41,16 @@ jobs:
     - name: Install package (including video extra for multimodal tests)
       run: $POETRY_HOME/bin/poetry install --extras video
     - name: Run tests with coverage
-      run: $POETRY_HOME/bin/poetry run pytest --cov=voyageai --cov-report=term-missing --cov-report=xml:coverage.xml --cov-fail-under=90
-    - name: Upload coverage report
+      run: $POETRY_HOME/bin/poetry run pytest --cov=voyageai --cov-report=term-missing --cov-report=xml:coverage.xml
+    - name: Enforce coverage threshold
+      # Gate once, on a single canonical version, instead of redundantly on
+      # every matrix entry. Threshold tracks the coverage this PR achieves.
       if: matrix.python-version == '3.12'
+      run: $POETRY_HOME/bin/poetry run coverage report --fail-under=93
+    - name: Upload coverage report
+      # always() so the report is still uploaded when the threshold gate fails,
+      # which is exactly when it's most useful for diagnosis.
+      if: always() && matrix.python-version == '3.12'
       uses: actions/upload-artifact@v4
       with:
         name: coverage-report
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,6 +13,11 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
+      # Run one Python version at a time: this job runs the full suite against
+      # the live API (incl. token-heavy voyage-multimodal-3.5), and running the
+      # whole matrix in parallel exceeds the project's per-minute token (TPM)
+      # rate limit, turning real assertions into spurious RateLimitError failures.
+      max-parallel: 1
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
     env:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,7 +41,7 @@ jobs:
     - name: Install package (including video extra for multimodal tests)
       run: $POETRY_HOME/bin/poetry install --extras video
     - name: Run tests with coverage
-      run: $POETRY_HOME/bin/poetry run pytest --cov=voyageai --cov-report=term-missing --cov-report=xml:coverage.xml
+      run: $POETRY_HOME/bin/poetry run pytest --cov=voyageai --cov-report=term-missing --cov-report=xml:coverage.xml --cov-fail-under=90
     - name: Upload coverage report
       if: matrix.python-version == '3.12'
       uses: actions/upload-artifact@v4
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -964,16 +964,31 @@ class TestVideoUtilsErrorPaths:
 
 
 class TestBaseClientConfig:
-    def test_get_client_config_valid(self):
+    def test_get_client_config_valid(self, tmp_path):
         from voyageai._base import _get_client_config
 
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -40,5 +40,18 @@ jobs:
...
-    - name: Run tests
-      run: $POETRY_HOME/bin/poetry run pytest
\ No newline at end of file
+    - name: Run tests with coverage
+      run: $POETRY_HOME/bin/poetry run pytest --cov=voyageai --cov-report=term-missing --cov-report=xml:coverage.xml
+    - name: Enforce coverage threshold
+      # Gate once, on a single canonical version, instead of redundantly on
+      # every matrix entry. Threshold tracks the coverage this PR achieves.
+      if: matrix.python-version == '3.12'
+      run: $POETRY_HOME/bin/poetry run coverage report --fail-under=93
```
jasonqiu95: **Critical — this gate will fail on every CI run.** It requires ≥93% total coverage, but this PR's own description reports **79% overall**, and a local no-API run measures ~66-67%. The live-API `test_client*.py` tests add more, but several modules stay far below 93% (`chunking.py` 17%, `client.py` 33%, `client_async.py` 31%, `object/*` largely untested), so the suite cannot reach 93%. `coverage report --fail-under=93` exits non-zero → the 3.12 matrix entry fails.

Root cause: the threshold was set to an aspirational number rather than the coverage the suite actually produces. It's also undocumented in the PR description, which frames this as a coverage-*visibility* change.

**Suggested fix:** set the threshold to a value the suite genuinely reaches, with headroom (confirmed against a real CI run), e.g.:
```yaml
        run: $POETRY_HOME/bin/poetry run coverage report --fail-under=75
```
Or drop the hard gate from this PR and introduce it in a follow-up once coverage is raised.

tests/test_unit.py
```diff
@@ -0,0 +1,979 @@
...
+            assert num_frames is not None
+            assert estimated_tokens is None
+
+
+# ---------------------------------------------------------------------------
+# _base — _get_client_config
+# ---------------------------------------------------------------------------
+
+
+class TestBaseClientConfig:
```
jasonqiu95, line 977: **These "unit" tests make real network calls.** The module docstring says these tests "do NOT call the Voyage API — they mock external dependencies," but `_get_client_config` calls `hf_hub_download(repo_id=f"voyageai/{model}", ...)`, so both tests hit HuggingFace Hub over the network. That makes them flaky (HF availability / rate limits) and slow, and breaks the file's stated contract.

**Suggested fix:** patch the download so the test is hermetic:
```python
with patch("voyageai._base.hf_hub_download", return_value="/tmp/client_config.json"):
    ...
```
(and for the invalid-model case, have the patched function raise the same error type HF raises).

tests/test_unit.py
```diff
@@ -0,0 +1,979 @@
...
+
+        config = _get_client_config("voyage-multimodal-3.5")
+        assert isinstance(config, dict)
+        assert "multimodal_image_pixels_min" in config
+
+    def test_get_client_config_invalid_model(self):
+        from voyageai._base import _get_client_config
+
+        with pytest.warns(match="Failed to load"):
+            with pytest.raises(Exception):
```
jasonqiu95: **`pytest.raises(Exception)` is too broad.** It passes for *any* exception — including one raised before the code under test — so it doesn't actually verify the `_get_client_config` failure path. `_get_client_config` does a bare `except: ... raise`, re-raising whatever `hf_hub_download` throws (an `HfHubHTTPError` / `RepositoryNotFoundError` for a missing repo).

**Suggested fix:** assert the concrete type, e.g.:
```python
from huggingface_hub.utils import HfHubHTTPError
with pytest.warns(match="Failed to load"), pytest.raises(HfHubHTTPError):
    _get_client_config("nonexistent-model-xyz-12345")
```

tests/test_unit.py
```diff
@@ -0,0 +1,979 @@
...
+# package (the `[video]` extra). Skip the whole class when either is missing so
+# the suite stays runnable without the extra installed instead of erroring.
+_FFMPEG_AVAILABLE = voyageai.video_utils.ffmpeg is not None and shutil.which("ffmpeg") is not None
+
+
+@pytest.mark.skipif(
+    not _FFMPEG_AVAILABLE,
+    reason="requires ffmpeg on PATH and the voyageai[video] extra",
+)
+class TestVideoUtilsIntegration:
```
jasonqiu95, line 788: **Redundancy / discoverability:** this integration class overlaps with the existing `tests/test_video_utils.py` (e.g. `test_optimize_video_e2e_example_video`, `_parse_fps`). Two video-test files exercising the same `optimize_video`/`_probe_video` paths makes it easy to add a case to one and miss the other. Consider consolidating the video tests into a single file.

jasonqiu95 (changes requested):
## Code Review: Add test coverage reporting to CI

**Summary:** The new `tests/test_unit.py` is a genuinely valuable, well-structured set of no-API unit tests, and the `voyageai/error.py` change is a real latent-bug fix. **Requesting changes on one blocking item:** the CI `--fail-under=93` gate is set *above* the coverage this PR actually achieves, so CI will hard-fail on every run. Details are in the inline comments.

### Critical
- **ci.yaml:49** — the 93% gate exceeds the coverage this PR reaches (PR says 79%; local no-API run ~66%). CI will always fail.

### Important
- **test_unit.py:966** — `TestBaseClientConfig` makes real HuggingFace Hub network calls despite the "no external API" docstring.
- **test_unit.py:978** — `pytest.raises(Exception)` is too broad.

### Suggestions
- **test_unit.py:788** — `TestVideoUtilsIntegration` overlaps with the existing `tests/test_video_utils.py`.

### Looks Good
- **error.py:63** — Real bug fix: old code referenced the nonexistent `api_resources.error_object.ErrorObject` module (would raise `AttributeError` on any structured error body). The lazy `VoyageResponse` import fixes it and breaks the import cycle. Nicely caught.
- `test_unit.py` overall is thorough and well-organized — good edge cases (None bodies, non-decodable bytes, proxy variants, pickle round-trip) and a clean `skipif` for the ffmpeg extra.

.github/workflows/ci.yaml
fzowl: Addressed in 56a7841. The gate is not aspirational in CI: the coverage job sets \`VOYAGE_API_KEY\` and runs the full \`pytest\`, so the live-API integration tests are included and the suite measures **93%** here (the latest run on this branch is green at exactly \`TOTAL 1360 97 93%\`). The ~66% / 79% figures were a no-API run / the stale PR description. I lowered the gate to \`--fail-under=90\` for headroom (~40 lines) so a single flaky API test or a small uncovered addition does not red the build, and documented this in the step comment.

tests/test_unit.py
fzowl: Fixed in 56a7841 — both \`TestBaseClientConfig\` tests now patch \`voyageai._base.hf_hub_download\`, so they no longer hit HuggingFace Hub and the file's "no external API" contract holds. The valid case reads a temp config file; the invalid case uses a mocked error.

tests/test_unit.py
fzowl: Fixed in 56a7841 — the invalid-model test now patches \`hf_hub_download\` to raise \`HfHubHTTPError\` and asserts \`pytest.raises(HfHubHTTPError)\` (inside the \`Failed to load\` warning), verifying the actual re-raise path instead of any exception.

tests/test_unit.py
fzowl: Fair point. I am leaving the consolidation as a separate follow-up rather than folding it into this PR: \`TestVideoUtilsIntegration\` here is intentionally part of the coverage-focused \`test_unit.py\`, and merging it into \`tests/test_video_utils.py\` is a pure test-reorg that is cleaner to review on its own. Tracking it as a cleanup item.

.github/workflows/ci.yaml
```diff
@@ -40,5 +40,18 @@ jobs:
...
+    - name: Upload coverage report
+      # always() so the report is still uploaded when the threshold gate fails,
+      # which is exactly when it's most useful for diagnosis.
+      if: always() && matrix.python-version == '3.12'
+      uses: actions/upload-artifact@v4
```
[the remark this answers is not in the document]
fzowl: Fixed in 56a7841 — pinned \`actions/upload-artifact\` (and \`actions/checkout\` / \`actions/setup-python\`) to full commit SHAs with version comments, removing the mutable-tag supply-chain risk.

fzowl:
Thanks @jasonqiu95. Addressed the review in 56a7841:

- **Coverage gate** — kept the gate but set `--fail-under=90`. The coverage job already sets `VOYAGE_API_KEY` and runs the full `pytest`, so the live-API integration tests are included and the suite measures **93%** in CI (latest run green). 90% leaves headroom so a flaky API test won't red the build; documented in the step comment.
- **Hermetic config tests** — `TestBaseClientConfig` now patches `voyageai._base.hf_hub_download` (no network), and the invalid-model case asserts the concrete `HfHubHTTPError` instead of a blanket `Exception`.
- **Supply chain** — pinned `actions/checkout`, `actions/setup-python`, and `actions/upload-artifact` to full commit SHAs.
- **Video test overlap** — left as a separate follow-up (pure test-reorg, cleaner to review on its own).

`tests/test_unit.py` (139 tests) passes locally.

tests/test_unit.py
```diff
@@ -0,0 +1,998 @@
...
+        from voyageai.video_utils import _compute_basic_usage_for_path
+
+        assert _compute_basic_usage_for_path("/nonexistent.mp4", model="x") == (None, None, None)
+
+    def test_get_video_token_config(self):
+        from voyageai.video_utils import _get_video_token_config
+
+        config = _get_video_token_config("voyage-multimodal-3.5")
+        assert config is not None
+        assert all(v > 0 for v in config)
```
jasonqiu95: This file's docstring promises "These tests do NOT call the Voyage API — they mock external dependencies", but `_get_video_token_config` routes through `voyageai._base._get_client_config`, which calls `hf_hub_download(repo_id="voyageai/voyage-multimodal-3.5", ...)` ([`_base.py#L33`](https://github.com/voyage-ai/voyageai-python/blob/99645d02c0594fce4d6e531f7468efa9eb64ef54/voyageai/_base.py#L29-L43)) — a live HuggingFace Hub fetch with no mock and no caching. So `assert config is not None` here only passes when the network is up; offline it returns `None` and the test fails. This is the same hermeticity gap you already fixed for `TestBaseClientConfig`, just missed in this class. The sibling tests `test_compute_basic_usage_for_path`, `test_video_from_file_no_optimize`, `test_optimize_video_from_path/_from_bytes/_from_video_object/_no_resize` make the same unmocked call (they tolerate `None`, so they don't fail offline, but they still hit the network).

Suggested fix — mirror the `TestBaseClientConfig` pattern and patch the download:

```suggestion
    def test_get_video_token_config(self, tmp_path):
        from voyageai.video_utils import _get_video_token_config

        config_file = tmp_path / "client_config.json"
        config_file.write_text(json.dumps({
            "multimodal_video_pixels_min": 100,
            "multimodal_video_pixels_max": 1000000,
            "multimodal_video_to_tokens_ratio": 100,
        }))
        with patch("voyageai._base.hf_hub_download", return_value=str(config_file)):
            config = _get_video_token_config("voyage-multimodal-3.5")
        assert config is not None
        assert all(v > 0 for v in config)
```

For the other tests, a class-level `autouse` fixture patching `voyageai._base.hf_hub_download` (or `voyageai.video_utils._get_video_token_config`) would keep the whole class hermetic.

jasonqiu95: This seems worth addressing since we are doing the same patch for TestBaseClientConfig

jasonqiu95 (approved):
Left a comment but feel free to merge after you addressed

PR #67, opened 2026-07-05
fix: forward input_type in async embedding helpers

NishchayMahor:
## What

The async embedding helpers validate `input_type` but then drop it. In `voyageai/embeddings_utils.py`:

- **Sync** `_get_embeddings` forwards it: `Embedding.create(input=..., model=..., input_type=input_type, **kwargs)`.
- **Async** `_aget_embeddings` validates it (`_check_input_type(input_type)`) but calls `Embedding.acreate(input=..., model=..., **kwargs)` — **`input_type` is a named parameter, not part of `**kwargs`, so it never reaches the API.**

Both public async functions (`aget_embedding`, `aget_embeddings`) route through `_aget_embeddings`, so:

```python
await aget_embedding(text, model="voyage-3", input_type="query")
```

computes the embedding with **no input-type prefix**, which degrades retrieval quality — a silent correctness bug and a sync/async divergence.

## Fix

Forward `input_type` in the async call, mirroring the sync path:

```python
await voyageai.Embedding.acreate(
    input=list_of_text, model=model, input_type=input_type, **kwargs
)
```

## Testing

Added `test_aget_embeddings_forwards_input_type` to `TestEmbeddingsUtils` (mirrors the existing `test_aget_embeddings_internal` mocked pattern) asserting `acreate` is called with `input_type="query"`. Fully offline (mocks `acreate`, no API key). It fails on `main` and passes with the fix.

### DIFF
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -623,9 +623,7 @@ class TestEmbeddingsUtils:
 
         mock_response = MagicMock()
         mock_response.data = [{"embedding": [0.1]}]
-        with patch(
-            "voyageai.Embedding.acreate", return_value=mock_response
-        ) as mock_acreate:
+        with patch("voyageai.Embedding.acreate", return_value=mock_response) as mock_acreate:
             await _aget_embeddings(["a"], model="voyage-3", input_type="query")
 
         assert mock_acreate.call_args.kwargs["input_type"] == "query"
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -617,6 +617,19 @@ class TestEmbeddingsUtils:
             result = await _aget_embeddings(["a", "b"], model="voyage-3")
             assert result == [[0.1], [0.2]]
 
+    @pytest.mark.asyncio
+    async def test_aget_embeddings_forwards_input_type(self):
+        from voyageai.embeddings_utils import _aget_embeddings
+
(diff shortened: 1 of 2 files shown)


fzowl (approved):
Verified locally with a live API key: full suite 370 passed / 21 skipped (local-model tests skip without local extras); the only failure was a transient `APIConnectionError` on an async multimodal case that moved between parametrizations across runs — environmental network flakiness, unrelated to this change. All embedding tests (the code path touched here) pass.

The fix is correct: async `_aget_embeddings` now forwards `input_type` to `acreate`, mirroring the sync path; since `input_type` is a named parameter it was previously silently dropped. No double-pass risk. The added offline test asserts the forwarded kwarg.

Note: the fork PR's Integration/`Tests` CI checks are red only because fork PRs don't receive the `VOYAGE_API_KEY` secret; with a key they pass (verified above).

fzowl:
@NishchayMahor Thank you very much for your contribution, great catch! I'm releasing a new version with this fix very soon!

NishchayMahor:
Thanks for the merge! Small fix, but the quiet kind that bites people at exactly the wrong time, so glad it's in. Big fan of the embedding models, retrieval quality is something we lean on a lot in the observability work I do. Happy to help out here again anytime.

PR #70, opened 2026-07-06
ci: fork-safe required checks + keyless coverage + drop needless max-parallel

### LINKED ISSUE #68: CI: drop needless max-parallel: 1 from unit-tests-no-local-deps job
Follow-up from review of #59 (thanks @jasonqiu95).

The `unit-tests-no-local-deps` job in `.github/workflows/ci.yaml` carries a `max-parallel: 1` and a comment claiming it "runs the full suite against the live API (incl. token-heavy voyage-multimodal-3.5)" and must serialize to stay under the TPM rate limit.

That justification is false for this job:
- No `VOYAGE_API_KEY` is set (workflow env has only `POETRY_HOME`; the only key is in the `integration-tests` job).
- Installs `--extras video` only (no local deps).
- Runs only `tests/test_client_local.py tests/test_client_local_async.py`.

With no key and no local deps, every `@pytest.mark.integration` / local-model test skips and the rest are offline validation / keyless-gate tests, so the job makes **zero** API calls. The comment/block was copied verbatim from `integration-tests` (lines 93-96), where it is correct.

### Effect
6-version matrix runs one Python at a time for no reason (~6x slower than needed); misleading comment.

### Fix
Drop `max-parallel: 1` and the copied comment from `unit-tests-no-local-deps`; keep serialization only on `integration-tests`.

```yaml
  unit-tests-no-local-deps:
    name: Unit Tests (no local deps)
    runs-on: ubuntu-22.04
    strategy:
      fail-fast: false
      matrix:
        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
```

Low priority (offline unit tests are fast), tracked separately so #59 can merge without a re-review cycle.

### LINKED ISSUE #69: CI: fork PRs can never pass required checks (integration tests need secrets forks don't get)
## Problem

Pull requests opened from **forks** can never satisfy the required status checks, so they are permanently `BLOCKED` from merging even when the code is correct.

Root cause: GitHub does **not** pass repository secrets to workflows triggered by `pull_request` from a fork (only `GITHUB_TOKEN` is provided). Approving the workflow run (the "Approve and run" gate for outside contributors) lets the workflow *execute* but still does **not** expose secrets. As a result:

- `integration-tests` has no `VOYAGE_API_KEY`, so its live-API tests fail on every fork PR.
- The legacy required-check bridge job `tests` (`name: Tests`, emitting the required `Tests (3.9)`..`Tests (3.12)` contexts) currently `needs: [unit-tests-no-local-deps, local-model-tests, integration-tests]`. Because it depends on `integration-tests`, the required `Tests (3.x)` contexts are red on forks too.

Net effect: no fork contribution can go green. This is not fixable by re-running.

Example: #67 (fork PR). The fix itself is correct — verified locally with a real API key: full suite 370 passed / 21 skipped, the only failure a transient `APIConnectionError` on an async multimodal case (network flakiness, unrelated). But the fork CI is red purely due to missing secrets.

## Recommended fix: required checks = fork-safe jobs only

Gate merges on checks that do **not** need secrets; run the secret-dependent suite where secrets exist.

1. `tests` bridge job `needs` only the fork-safe jobs: `[unit-tests-no-local-deps, local-model-tests]`. Then the required `Tests (3.x)` contexts can go green on forks.
2. `integration-tests`: skip on fork PRs (run on `push` to `main` and on same-repo PRs, where secrets are available), e.g. `if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository`. Integration coverage then runs post-merge on `main`; a regression is caught there and can be reverted.
3. Branch protection (admin): ensure the required set is the fork-safe checks (lint, `Unit Tests (no local deps) (3.x)`, `Local Model Tests (3.x)`, and the `Tests (3.x)` bridge). `Integration Tests (3.x)` should **not** be required.

## Related
- #68 (drop the stale `max-parallel: 1` / update required check names to the new job names). The required-check rename there and the fork-safety change here should be reconciled together.

fzowl:
Fixes #69 and closes #68 — CI-workflow-only changes.

## #69 — fork PRs can never pass required checks
GitHub does not pass repository secrets to `pull_request` workflows from a fork, so `integration-tests` (needs `VOYAGE_API_KEY`) is always red on fork PRs. The legacy required-check bridge `tests` depended on it, so the required `Tests (3.x)` contexts were red on every fork PR — permanently `BLOCKED`.

- **`tests` bridge `needs` only fork-safe, keyless jobs** → required `Tests (3.x)` can go green on forks.
- **`integration-tests` runs everywhere** (no skip); red on fork PRs by design, gated by **review** (external: merge + revert if the post-merge `main` run fails; same-repo: don't merge on red). Not a required check (confirmed in the merge UI).

## Keyless required coverage
The keyless mock suite (`test_unit.py`, `test_import_order.py`) previously ran only inside the non-required `integration-tests`, so a same-repo PR breaking `test_unit.py` would pass every required check. Now covered by the required feeder `unit-tests-no-local-deps` (no API key). `pytest -m "not integration"` was rejected — it is not keyless (the live-API client/multimodal tests carry no `integration` marker; 147 fail without a key).

**`test_video_utils.py` is deliberately NOT in the required jobs:** its e2e leg calls `_get_client_config()` → `hf_hub_download()` with no offline fallback, so it needs the HuggingFace Hub and would make a required check flaky on HF outages. It runs in `integration-tests`.

## Matrix isolation
`needs` is job-level, so a failing 3.13/3.14 leg of a bridge feeder would fail the whole job and skip the bridge → required contexts unreported → blocked. Bridge feeders are the required set **3.9-3.12**; **3.13/3.14 run in separate advisory jobs** (`*-next`): red on failure (visible) but not bridge dependencies, so they never block merges (rerun turns them green).

## Dedup
The four test jobs shared an identical install/run block. Extracted into a local composite action (`.github/actions/setup-and-test`: `python-version`/`extras`/`pytest-args`/`install-openblas`). Four separate jobs and the `needs` topology kept intact (no `matrix.include`+`continue-on-error` collapse, which would re-couple non-required versions to the required gate).

## #68 — needless `max-parallel: 1` on `unit-tests-no-local-deps`
Zero API calls (no key, no local extras) → the TPM rationale copied from `integration-tests` doesn't apply. Removed; serialization stays only on `integration-tests`.

## Admin follow-up
None. The `Tests` bridge keeps emitting the legacy required contexts; `Integration Tests (3.x)` is not required. The bridge can be retired later if protection is switched to require the real job names.

Related: #67 (the fork PR that surfaced this).


.github/workflows/ci.yaml
```diff
@@ -143,24 +148,25 @@ jobs:
...
-  # This thin aggregate job runs no tests; it only re-emits the required
-  # "Tests (<pyver>)" contexts and passes iff all three real jobs pass, so the
-  # required check still reflects a green suite. Remove this job once an admin
-  # updates branch protection to the new job names (tracked in #68).
+  # Remove this job once an admin updates branch protection to the new job names
+  # (tracked in #68).
   tests:
     name: Tests
-    needs: [unit-tests-no-local-deps, local-model-tests, integration-tests]
+    needs: [unit-tests-no-local-deps, local-model-tests]
```
jasonqiu95, line 169: **Required checks no longer cover the keyless unit tests.** These two bridge deps only run `test_client_local.py` + `test_client_local_async.py`. The pure/mock tests (`test_unit.py`, `test_client.py`, `test_client_async.py`, `test_client_multimodal.py`, `test_import_order.py`, `test_video_utils.py`) run **only** inside `integration-tests` (bare `pytest`, no path/marker filter) — which this PR deliberately makes non-required. Net effect: a same-repo PR that breaks `test_unit.py` (needs no API key) now passes every required `Tests (3.x)` check; before this PR it did not.

**Suggested fix:** add a fork-safe job that runs the keyless suite with no `VOYAGE_API_KEY` (e.g. `pytest -m "not integration"`, or the explicit mock/unit file list) and include it in the bridge's `needs`:
```yaml
needs: [unit-tests-no-local-deps, local-model-tests, unit-tests-mock]
```

.github/workflows/ci.yaml
```diff
@@ -143,24 +148,25 @@ jobs:
...
     runs-on: ubuntu-22.04
     strategy:
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12"]
```
jasonqiu95, line 173: **Matrix mismatch can still leave fork PRs BLOCKED — the exact thing this PR fixes.** `needs` is job-level, not per-matrix-leg. The bridge matrix is 3.9–3.12, but `unit-tests-no-local-deps` / `local-model-tests` run 3.9–**3.14**. If either fails on its **3.13 or 3.14** leg (a version outside the required set), the *whole* upstream job is marked failed → the `tests` bridge is **skipped** → the required `Tests (3.9)…(3.12)` contexts never report → PR stuck BLOCKED, unfixable by rerun. `fail-fast: false` doesn't help: it keeps sibling legs running but the job still reports failure.

**Suggested fix:** don't let an advisory 3.13/3.14 leg fail a required-gating job — split 3.13/3.14 into a separate non-required job, or mark those legs `continue-on-error: true`, so the required 3.9–3.12 legs always determine the bridge.

.github/workflows/ci.yaml
```diff
@@ -143,24 +148,25 @@ jobs:
...
+  # "Tests (<pyver>)" contexts can go green on fork PRs. integration-tests is
+  # deliberately excluded: it needs secrets that fork PRs never receive, and is
+  # covered post-merge by the push-to-main run instead (#69).
   #
-  # This thin aggregate job runs no tests; it only re-emits the required
-  # "Tests (<pyver>)" contexts and passes iff all three real jobs pass, so the
-  # required check still reflects a green suite. Remove this job once an admin
-  # updates branch protection to the new job names (tracked in #68).
+  # Remove this job once an admin updates branch protection to the new job names
+  # (tracked in #68).
```
jasonqiu95: **Stale tracking reference.** This says the branch-protection migration is *tracked in #68* — but #68 is the *max-parallel* issue, and this PR `closes #68`. So the follow-up now points at a closed, unrelated issue (and #69 is closed by this PR too). The one-time admin step described in the PR body has no open tracker.

**Suggested fix:** file a dedicated "update branch protection to new job names" issue and reference it here instead of #68.

jasonqiu95 (commented):
Automated review of the CI changes. Three points on the required-check bridge; all are on the workflow semantics, none block the fork-safe intent conceptually.

One extra note (not inline, pre-existing): `local-model-tests` is a required bridge dependency but is not hermetic — it downloads `voyage-4-nano` from HuggingFace with no `-m` filter, so a Hub outage/anonymous-throttle can red a required check on fork PRs. Same job-level cascade as the matrix note below.

.github/workflows/ci.yaml
fzowl: Agreed on the coverage gap — fixed. But `pytest -m "not integration"` is not keyless: only test_client_local*.py carry the integration marker; the live-API client/multimodal tests don't, so `not integration` includes them and 147 fail without a key (verified locally). Instead the genuinely-keyless files now run in the required feeder `unit-tests-no-local-deps` (test_unit.py + test_video_utils.py + test_import_order.py, plus the local-import files), all with no API key, feeding the required Tests bridge. Verified green keyless: 178 passed / 21 skipped.

.github/workflows/ci.yaml
fzowl: Agreed — fixed. The bridge feeders (`unit-tests-no-local-deps`, `local-model-tests`) are now the required set 3.9-3.12 only, matching the bridge matrix, so no non-required leg can fail the job and skip the bridge. 3.13/3.14 moved to separate advisory jobs (`*-next`): red on failure (kept visible) but not bridge dependencies, so they never block merges — a rerun turns them green.

.github/workflows/ci.yaml
fzowl: Fixed. Removed the issue references from the code comment (it's self-contained now). No branch-protection rename is needed: the Tests bridge keeps emitting the legacy `Tests (3.x)` contexts, and `Integration Tests (3.x)` is not a required check (confirmed in the merge UI), so no admin migration is pending. PR body updated accordingly.

.github/workflows/ci.yaml
```diff
@@ -46,17 +52,96 @@ jobs:
         ffmpeg -version
     - name: Install package (with video extra, without local extras)
       run: $POETRY_HOME/bin/poetry install --extras video
-    - name: Run unit tests (no API key, no local deps)
-      run: $POETRY_HOME/bin/poetry run pytest tests/test_client_local.py tests/test_client_local_async.py -v
+    - name: Run keyless unit tests (no API key, no local deps)
+      run: >-
+        $POETRY_HOME/bin/poetry run pytest
+        tests/test_unit.py tests/test_video_utils.py tests/test_import_order.py
```
jasonqiu95: **Adding `test_video_utils.py` to this required, "zero API calls" job makes a required check depend on the HuggingFace Hub.** `test_optimize_video_e2e_example_video` (a `nullcontext` happy-path leg) calls `_get_client_config("voyage-multimodal-3.5")` directly (`tests/test_video_utils.py:211`), which calls `hf_hub_download(...)` in `_base.py` with no offline fallback and no mock. Fresh runners have an empty HF cache, so this hits the network. An HF outage / anonymous rate-limit / repo gating then reds a **required** `Tests (3.x)` context on every PR (incl. forks) — no code broken, no secret missing. The header comment on line 14 ("Makes zero API calls, so it is fork-safe") overclaims.

**Suggested fix:** make the test hermetic the same way `test_unit.py::TestBaseClientConfig` already does — patch the download to a local fixture:
```python
with patch("voyageai._base.hf_hub_download", return_value=str(config_file)):
    cfg = _get_client_config("voyage-multimodal-3.5")
```
Or keep this e2e test out of the fork-safe job (it's really an integration test). Either restores the "zero API calls" invariant the comment promises.

.github/workflows/ci.yaml
```diff
@@ -46,17 +52,96 @@ jobs:
...
+        tests/test_client_local.py tests/test_client_local_async.py -v
+
+  # Advisory: same keyless suite on newer Python versions not in the required
+  # set. Red on failure (visible), but not a dependency of the "Tests" bridge,
+  # so a 3.13/3.14 break does not block merges; a rerun turns it green.
+  unit-tests-no-local-deps-next:
```
jasonqiu95, line 53: **The `-next` jobs are near-verbatim copies of their required counterparts** — `unit-tests-no-local-deps-next` vs `unit-tests-no-local-deps` are byte-identical in `steps:` (26/26 lines), as are the two local-model jobs (25/25); they differ only in the matrix versions and name. Install steps (ffmpeg, OpenBLAS, poetry install, the pytest path list) must now be edited in 4 places, inviting drift.

Note: the obvious `matrix.include` + `continue-on-error` collapse is **not** a safe fix here — folding 3.13/3.14 back into a job the bridge `needs:` re-couples non-required versions to the required contexts, and whether a `continue-on-error` matrix leg leaves the job's `result` as `success` for `needs:` is not documented (the `outcome`→`conclusion` upgrade is specified for *steps*, not matrix legs). That would risk re-breaking the exact gate this PR fixed.

**Suggested fix:** keep the four separate jobs and their `needs:` topology, but extract the shared steps into a local composite action (`.github/actions/setup-and-test` taking `python-version`, `extras`, `pytest-args`, `install-openblas` inputs). Each job shrinks to a couple of lines calling it — dedup without touching gate semantics. Or just leave as-is; it's safe, only repetitive.

jasonqiu95 (commented):
Re-reviewed after `c9f78a7`. All three prior points are addressed correctly:

- **Coverage gap** — fixed. And your pushback is right: `contextualized_embed` calls `_require_api_key()` *before* validation (`client.py:129`), and `AuthenticationError` is a `VoyageError`, not a `ValueError` — so `test_client.py`'s validation tests would raise `AuthenticationError` without a key and can't be keyless. No residual gap there; the genuinely-keyless files are the right set.
- **Matrix/bridge decoupling** — fixed. Feeders and bridge are both 3.9–3.12; 3.13/3.14 moved to non-required `*-next` jobs. Correct.
- **Stale #68 ref** — fixed.

Two new points that the fix commit itself introduced (one is a real flakiness risk on a required check):

.github/workflows/ci.yaml
fzowl: Good catch — confirmed and fixed. `_get_client_config()` in _base.py calls `hf_hub_download()` with no offline fallback, and `test_optimize_video_e2e_example_video` calls it unmocked, so it hits the HF Hub; my local run passed only because the cache was warm. Removed `test_video_utils.py` from the fork-safe jobs (it's effectively integration: ffmpeg + HF), so the required jobs are hermetic again. It still runs in integration-tests. Required keyless coverage stays via test_unit.py (which mocks the HF download) + test_import_order.py. Updated the job comment to stop overclaiming.

.github/workflows/ci.yaml
fzowl: Agreed. Extracted the shared install/run block into a local composite action (`.github/actions/setup-and-test` with python-version/extras/pytest-args/install-openblas inputs); each job is now ~5 lines. Kept the four separate jobs and the `needs` topology exactly as-is — did not use the matrix.include + continue-on-error collapse, for the reason you noted (it would re-couple non-required versions to the required contexts, and the matrix-leg result semantics for `needs` aren't documented). Net -40 lines.

fzowl:
Self-review hardening (pre-empting likely comments):

- **HF flakiness in the required gate.** `local-model-tests` loads the model via `SentenceTransformerBackend(...)`, which downloads from the HF Hub — same outage exposure as the `test_video_utils` e2e that was pulled out, but here the fixture deliberately lets download errors surface (documented) rather than skip-masking bugs. Rather than weaken that, added an `actions/cache` of `~/.cache/huggingface` on the model-loading jobs; huggingface_hub falls back to a populated cache when the Hub is unreachable, so routine HF flake no longer reds the required check.
- **Action input handling.** The composite now passes `extras`/`pytest-args` via `env:` instead of interpolating `${{ inputs.* }}` into `run:` (script-injection pattern actionlint flags).
- **CI cost.** Added a `concurrency` group so superseded PR runs cancel; not cancelled on `main` so each merged commit is still validated post-merge.

Consciously deferred (happy to take if wanted): single-source `POETRY_HOME` (composite hardcodes `/opt/poetry` vs the workflow `env`); coverage `--fail-under` only runs in the non-required `integration-tests` so it isn't enforced on fork PRs; `pytest-args` duplicated across the required/`-next` pairs.

jasonqiu95 (commented):
Re-reviewed after `43b9d90` + `6d57695`. Both round-2 findings are resolved, and I checked the new surface for regressions:

**1. HF network in a required check — resolved (and verified deeper than the fix comment claims).** `test_video_utils.py` is out of the fork-safe jobs. Note that `test_unit.py`'s `TestVideoUtilsIntegration` *stays* in the required job and still calls `optimize_video` / `_compute_basic_usage_for_path` / `Video.from_file` with `model="voyage-multimodal-3.5"` **unmocked** — so it's not "mocked" as the reply says. But it's genuinely safe: `_get_video_token_config` swallows HF errors → `None` (`video_utils.py:308-312`), both `optimize_video` and `_compute_basic_usage_for_path` have `video_config is None` fallbacks, and no test in that class asserts on the config-derived `estimated_tokens` — they only assert ffmpeg-probe values (`num_pixels`, `num_frames`, `.optimized`). So an HF outage degrades gracefully instead of reding the required check. Confirmed resolved.

**2. Job duplication — resolved.** The composite action (`.github/actions/setup-and-test`) collapses the four jobs to ~5 lines each, keeps the `needs`/job topology intact, and passes inputs via `env:` (not `${{ }}` interpolation) — good injection hygiene. `POETRY_HOME` is hardcoded in the action (correct — composite actions don't inherit workflow `env`) and the only workflow-level `$POETRY_HOME` consumer is the coverage step, which does inherit it. No mismatch.

**New additions checked, all correct:**
- `concurrency`: `github.ref` is `refs/pull/<N>/merge` (unique per PR), and `cancel-in-progress` is gated to `pull_request` so the post-merge `main` integration gate is never cancelled.
- HF `actions/cache`: fixed key across parallel legs → save races are non-fatal warnings; warm cache lets `huggingface_hub` fall back on an outage, a strict improvement over the prior always-live behavior. (Residual: a cold cache + HF outage can still red the required `local-model-tests` feeder, but that's pre-existing, not introduced here.)
- `PYTEST_ARGS` word-splitting is safe for all three arg sets (no spaces/globs); `SC2086` disable is appropriate. Checkout precedes the local action in every job; all actions SHA-pinned.

No new correctness issues. LGTM from my side.

jasonqiu95 (approved):

